PR #4, opened 2018-07-18
Find project by ID

robinlacey:
**What** 
Find a project by passing 'id' param to /project/find with GET

**Why**
GET /project/find uses ProjectGateway find_by method. Created a use case for FindProject rather than calling project_gateway find_by direct


spec/acceptance/homes_england/get_hif_project_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  after do
+    File.delete(ENV['PROJECT_FILE_PATH'])
+  end
+
+  it 'should find a project by its id' do
+    response = get_use_case(:create_new_project).execute(type: '', baseline: {})
+    project = get_use_case(:find_project).execute(id: response[:id])
+    expect(project).to_not be_nil
```
robinlacey: kept this light as functionality is already being used in create. could wrap with more tests on data out?

PR #5, opened 2018-07-18
GET Project by Id

robinlacey:
GET : /project/find with id param will find Project
Added FindProject use case

### DIFF
--- a/spec/acceptance/homes_england/get_hif_project_spec.rb
+++ b/spec/acceptance/homes_england/get_hif_project_spec.rb
@@ -1,0 +1,21 @@
+require 'rspec'
+require_relative '../shared_context/use_case_factory'
+
+describe 'Getting a HIF project' do
+  include_context 'use case factory'
+
+  before do
+    ENV['PROJECT_FILE_PATH'] = '/tmp/projects.json'
+    File.open(ENV['PROJECT_FILE_PATH'], 'w') {}
+  end
+
+  after do
+    File.delete(ENV['PROJECT_FILE_PATH'])
+  end
+
+  it 'should find a project by its id' do
+    response = get_use_case(:create_new_project).execute(type: '', baseline: {})
+    project = get_use_case(:find_project).execute(id: response[:id])
+    expect(project).to_not be_nil
+  end
+end
--- a/spec/acceptance/homes_england/get_hif_project_spec.rb
+++ b/spec/acceptance/homes_england/get_hif_project_spec.rb
@@ -14,8 +14,11 @@ describe 'Getting a HIF project' do
   end
 
   it 'should find a project by its id' do
-    response = get_use_case(:create_new_project).execute(type: '', baseline: {})
+    response = get_use_case(:create_new_project).execute(type: 'hif',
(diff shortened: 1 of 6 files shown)


spec/unit/homes_england/use_case/find_project_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'rspec'
+
+describe HomesEngland::UseCase::FindProject do
```
danielburnley: This spec probably needs a few more cases, a couple examples for triangulation as well as needing some changes as a result of the above change (no longer returning domain objects, instead returning a hash)

Might also be worthwhile potentially considering what happens if the project doesn't exist? The use case could return nil, or it could return a hash containing an error, not sure what the best would be

danielburnley: I'd probably recommend tests like:

```
example 1
  finds the id on the gateway
  returns a hash containing the project type
  returns a hash containing the project data
example 2
  same as above
```

spec/unit/homes_england/use_case/find_project_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+  let(:project_gateway) { double(find_by: project) }
+  let(:use_case) { described_class.new(project_gateway: project_gateway) }
+  let(:response) { use_case.execute(id: id) }
+
+  context 'given an id' do
+    let(:project) { { cats: 'meow' } }
+    let(:id){42}
+    it 'gateway find_by method is used' do
```
danielburnley: This should probably assert something like: 

`expect(project_gateway).to have_recieved(:find_by).with(id: id)`

spec/web_routes/finding_project_spec.rb
```diff
@@ -0,0 +1,91 @@
+require 'rspec'
+require_relative 'delivery_mechanism_spec_helper'
+
+describe 'Finding a project' do
```
danielburnley: Be worth going over this one in person to explain a bit more about Rack tests and remove some of the stuff that's unnecessary/clear some stuff up - it's a pretty solid first pass through just a few things that aren't immediately clear when using this framework

spec/acceptance/homes_england/get_hif_project_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    ENV['PROJECT_FILE_PATH'] = '/tmp/projects.json'
+    File.open(ENV['PROJECT_FILE_PATH'], 'w') {}
+  end
+
+  after do
+    File.delete(ENV['PROJECT_FILE_PATH'])
+  end
+
+  it 'should find a project by its id' do
+    response = get_use_case(:create_new_project).execute(type: '', baseline: {})
```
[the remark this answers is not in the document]
robinlacey: I didn't know how far to go with this spec as it was essentially testing existing functionality.  Would be good to go over tomorrow. Thought it was better to have something over nothing!

danielburnley:
Changes have been addressed in #6

PR #6, opened 2018-07-19
POST Update Project

robinlacey:
Posting to /project/update will update the project via the gateway. Currently posting with json:
`id: <id>,
project: {
    type: <type>,
    baselineData: <data>
}`
Will return a `404` if `id` is not found or `project`/`id` is `nil`. (Will add additional checks once we have worked out the validation.)

### DIFF
--- a/lib/delivery_mechanism/web_routes.rb
+++ b/lib/delivery_mechanism/web_routes.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require 'sinatra'
 
 module DeliveryMechanism
@@ -7,7 +9,7 @@ module DeliveryMechanism
     end
 
     get '/project/find' do
-      return 404 if request.env['rack.request.query_hash']['id'].nil?
+      return 404 if params['id'].nil?
 
       return_project = @use_case_factory.get_use_case(:find_project).execute(id: params['id'].to_i)
 
@@ -41,16 +43,16 @@ module DeliveryMechanism
 
       if valid_update_request_body(request_body)
         use_case = @use_case_factory.get_use_case(:update_project)
-        success = use_case.execute(
+        update_successful = use_case.execute(
           id: request_body['id'],
           project: {
             type: request_body['project']['type'],
             baseline: request_body['project']['baselineData']
           }
-        )
-        response.status = success[:success] != 'true' ? 404 : 200
+        )[:success]
+        response.status = update_successful ? 200 : 404
       else
-        response.status = 404
+        response.status = 400
       end
     end
 
--- a/lib/delivery_mechanism/web_routes.rb
+++ b/lib/delivery_mechanism/web_routes.rb
@@ -6,6 +6,13 @@ module DeliveryMechanism
   class WebRoutes < Sinatra::Base
     before do
       @use_case_factory = Dependencies.use_case_factory
+      response.headers["Access-Control-Allow-Origin"] = '*'
+    end
+
+    options '*' do
+      response.headers["Access-Control-Allow-Origin"] = '*'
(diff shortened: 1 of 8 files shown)


lib/delivery_mechanism/web_routes.rb
```diff
@@ -38,6 +36,30 @@ class WebRoutes < Sinatra::Base
       response.status = 200
     end
 
+    post '/project/update' do
+      request_body = JSON.parse(request.body.read)
+
+      if valid_update_request_body(request_body)
```
danielburnley: Would it be 404 if it's an invalid request? I'd expect 404 if the project didn't exist but.. the other one if it was invalid

lib/delivery_mechanism/web_routes.rb
```diff
@@ -38,6 +36,30 @@ class WebRoutes < Sinatra::Base
...
+        use_case = @use_case_factory.get_use_case(:update_project)
+        success = use_case.execute(
+          id: request_body['id'],
+          project: {
+            type: request_body['project']['type'],
+            baseline: request_body['project']['baselineData']
+          }
+        )
+        response.status = success[:success] != 'true' ? 404 : 200
```
danielburnley: `success[:success]` eh? You could put the `[:success]` after the `usecase.execute` brackets to make this read a little less weird

lib/homes_england/gateway/file_project.rb
```diff
@@ -20,15 +20,31 @@ def create(project)
     projects.length - 1
   end
 
+  def update(id:, project:)
+    if id.nil? || project.nil? || saved_projects.nil? || saved_projects[id].nil?
```
danielburnley: Could make this read better, `if project_exists?`

Additionally we don't really need the defensive programming as the usec ases that call this will always pass the correct stuff in, or at least, we trust developers to always pass this stuff in correctly. Validation on id/project will be done further up the tree

lib/homes_england/gateway/file_project.rb
```diff
@@ -20,15 +20,31 @@ def create(project)
...
+      { success: 'false' }
+    end
+    projects = saved_projects
+    projects[id] = {
+      type: project[:type],
+      data: project[:baseline]
+    }
+    File.open(@file_path, 'w') do |f|
```
danielburnley: could probably move this out into a method as it's probably repeated

lib/homes_england/gateway/file_project.rb
```diff
@@ -57,9 +72,9 @@ def deep_symbolize_keys(obj)
       end
       result
     when Array
-      obj.map {|value| deep_symbolize_keys(value)}
+      obj.map { |value| deep_symbolize_keys(value) }
```
danielburnley: we can delete this entire method whoops, forgot to do that

lib/delivery_mechanism/web_routes.rb
robinlacey: Switched to 400

lib/delivery_mechanism/web_routes.rb
robinlacey: switched to bools for the time being, a bit easier on the eye

lib/homes_england/gateway/file_project.rb
robinlacey: 👍 done

lib/homes_england/gateway/file_project.rb
robinlacey: Moved to method. I agree this should be cleaned up upstream - will remove when we have this in place.

lib/homes_england/gateway/file_project.rb
robinlacey: done 💯

PR #9, opened 2018-07-25
Create and Get Return

robinlacey:


lib/homes_england/use_case/create_return.rb
```diff
@@ -0,0 +1,9 @@
+class HomesEngland::UseCase::CreateReturn
+  def initialize(return_gateway:)
+    @return_gateway = return_gateway
+  end
+
+  def execute(*args)
```
danielburnley: for future reference you can just put `*`

spec/unit/homes_england/use_case/get_return_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+require 'rspec'
+
+describe HomesEngland::UseCase::GetReturn do
+  let(:return_gateway){spy(get_return: {cats: "Catslol"}) }
+  let(:get) { described_class.new(return_gateway: return_gateway) }
+  it 'can be initialised' do
+    described_class.new(return_gateway: return_gateway)
+  end
+
+  it 'will run get_return on return_gateway with parameters' do
```
danielburnley: `it 'will pass the correct id to the return gateway`? 

Additionally there's no triangulation in this test, so it could pass with just a hardcoded `0`

lib/homes_england/use_case/create_return.rb
```diff
@@ -0,0 +1,9 @@
+class HomesEngland::UseCase::CreateReturn
```
danielburnley: `LocalAuthority::UseCase::CreateReturn` ? The Local Authorities create returns - not Homes England

This isn't too vital, but it's worth keeping in mind that we should structure this based on the actors

( Your gateway would then potentially live in `Common` as it's tied to both :thinking: not actually sure where it'd live)

spec/acceptance/homes_england/perform_return_spec.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+require 'rspec'
+require_relative '../shared_context/use_case_factory'
+
+describe 'Performing Return on HIF Project' do
+  include_context 'use case factory'
+
+  before do
```
danielburnley: Getting rid of this will be so good

spec/unit/homes_england/use_case/get_return_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    described_class.new(return_gateway: return_gateway).execute(id: 0)
+    expect(return_gateway).to have_received(:get_return).with(id: 0)
+  end
+
+  context 'example one' do
+    let(:return_gateway){spy(get_return: {cats: "Catslol"}) }
```
danielburnley: Currently, this means that our use case is leaking domain objects from the gateway - which causes our delivery mechanism is now tightly coupled to the domain object 

You should always communicate between gateways/use cases with domain objects - so in this case i'd expect the spy gateway to return a domain object instead of this hash

Your use case would then transform this domain object into a hash for it to return

spec/unit/homes_england/use_case/get_return_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    it 'will return the object from get_return' do
+      return_data = described_class.new(return_gateway: return_gateway).execute(id: 0)
+      expect(return_data).to eq({cats: "Catslol"})
```
danielburnley: I'd expect this to more look more like: 

```
{ return_data: { foo: 'bar' } }
```

Or something like that - some data structure that our delivery mechanism can transform

_or_ we could start introducing presenters now - but just a hash will work

spec/unit/homes_england/use_case/get_return_spec.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+require 'rspec'
+
+describe HomesEngland::UseCase::GetReturn do
+  let(:return_gateway){spy(get_return: {cats: "Catslol"}) }
+  let(:get) { described_class.new(return_gateway: return_gateway) }
+  it 'can be initialised' do
```
danielburnley: Can probably delete this test as it gets initialised elsewhere

spec/unit/homes_england/use_case/get_return_spec.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+require 'rspec'
+
+describe HomesEngland::UseCase::GetReturn do
+  let(:return_gateway){spy(get_return: {cats: "Catslol"}) }
+  let(:get) { described_class.new(return_gateway: return_gateway) }
```
danielburnley: `let(:use_case)` ? Reads a bit nicer than get

spec/unit/homes_england/use_case/create_return_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+require 'rspec'
+
+describe HomesEngland::UseCase::CreateReturn do
+  let(:return_gateway) { spy(create: 1) }
+  let(:create_return) { described_class.new(return_gateway: return_gateway) }
+  it 'can be created' do
+    described_class.new(return_gateway: return_gateway)
+  end
+
+  it 'can be executed with an id' do
```
danielburnley: This can be deleted as it's tested elsewhere

spec/unit/homes_england/use_case/create_return_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    create_return.execute(id: 0, return_data: {})
+  end
+
+  it 'creates a return via the gateway' do
```
danielburnley: I'd expect this to be split into a couple of tests (both living within the examples):

`it 'passes the id to the gateway'`
`it 'passes the return to the gateway'`

The second on asserting that the correct domain object is passed

lib/homes_england/use_case/create_return.rb
```diff
@@ -0,0 +1,9 @@
...
+  def initialize(return_gateway:)
+    @return_gateway = return_gateway
+  end
+
+  def execute(*args)
+    @return_gateway.create(*args)
```
danielburnley: This should call the gateway with a domain object, not just a raw hash

lib/homes_england/use_case/get_return.rb
```diff
@@ -0,0 +1,9 @@
+class HomesEngland::UseCase::GetReturn
+  def initialize(return_gateway:)
+    @return_gateway = return_gateway
+  end
+
+  def execute(*args)
+    @return_gateway.get_return(*args)
```
danielburnley: (as mentioned below) this should be transforming a domain object into a hash

spec/unit/homes_england/use_case/get_return_spec.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+require 'rspec'
+
+describe HomesEngland::UseCase::GetReturn do
```
danielburnley: What happens if the Return doesn't exist? probably want to add a test for the gateway returning `nil`

lib/homes_england/use_case/create_return.rb
robinlacey: Good spot - moved to Local Authority

lib/local_authority/use_case/create_return.rb
```diff
@@ -0,0 +1,9 @@
+class LocalAuthority::UseCase::CreateReturn
+  def initialize(return_gateway:)
+    @return_gateway = return_gateway
+  end
+
+  def execute(return_domain_in)
```
danielburnley: This should also receive a hash instead of a domain object - Otherwise the place calling it is coupled to the domain object

lib/local_authority/use_case/get_return.rb
```diff
@@ -0,0 +1,10 @@
+class LocalAuthority::UseCase::GetReturn
+  def initialize(return_gateway:)
+    @return_gateway = return_gateway
+  end
+
+  def execute(*args)
```
danielburnley: Should properly name your arguments here for readability

lib/local_authority/use_cases.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class LocalAuthority::UseCases
+  def self.register(builder)
+    builder.define_use_case :project_gateway do
+      LocalAuthority::Gateway::FileProject.new(
```
danielburnley: Shouldn't this use the Project gateway we already have? In which case you don't need to define it again

spec/acceptance/homes_england/perform_return_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+      },
+      financial: {
+        date: '2017-12-25',
+        funded_through_HIF: true
+      }
+    }
+
+    project = get_use_case(:create_new_project).execute(type: 'hif', baseline: project_baseline)
+
+    create_return_one_data = {
```
danielburnley: you could reduce this down to:

```ruby
project_baseline[:summary][:status] = 'Cats cats cats'
```

but also this is more clear as to what you're passing in so i might actually like yours better for an acceptance spec

spec/unit/local_authority/use_case/create_return_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+
+describe LocalAuthority::UseCase::CreateReturn do
+  let(:return_to_create) { double(project_id: 0, data: { summary: { name: 'Cats' } }) }
+  let(:return_gateway) { spy(create: {}) }
+  let(:create_return) { described_class.new(return_gateway: return_gateway) }
+  it 'can be created' do
+    described_class.new(return_gateway: return_gateway)
+  end
+
+  context 'creates a return via the gateway' do
```
danielburnley: There should be one of these in each example, currently you could make this test pass with a hardcoded implementation

danielburnley:
```
example one
  passes the id
  passes the return
  returns the created id
example two
  passes id
  passes return
  returns the id
```

lib/local_authority/use_case/create_return.rb
```diff
@@ -0,0 +1,9 @@
...
+    @return_gateway.create(return_domain_in)
```
danielburnley: be nicer for this to return `{ id: <id> }` imo - so we're always returning hashes

spec/unit/local_authority/use_case/get_return_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  let(:return_hash) do
+    {
+      project_id: 0,
+      data:  { cats: 'Jerry' }
+    }
+  end
+
+  let(:use_case) { described_class.new(return_gateway: return_gateway).execute(id: 0) }
```
danielburnley: I'd call this `response` - because it's actually set to the response of the use case (usecases operating on request/responses)

spec/unit/local_authority/use_case/get_return_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  it 'will pass the correct id to the return gateway' do
+    use_case
```
danielburnley: you could put `before { response }` to avoid having to call use case all the time

spec/unit/local_authority/use_case/get_return_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    expect(return_gateway).to have_received(:get_return).with(id: 0)
+  end
+
+  context 'example one' do
+    it 'will return the object from get_return' do
+      return_hash = use_case
```
danielburnley: setting it to response makes this a bit nicer to read

```
response[:project_id]
response[:data]
```

spec/unit/local_authority/use_case/get_return_spec.rb
faissalMT: I've put something similar since your example would only work for one test.

lib/local_authority/use_case/create_return.rb
```diff
@@ -0,0 +1,14 @@
+class LocalAuthority::UseCase::CreateReturn
+  def initialize(return_gateway:)
+    @return_gateway = return_gateway
+  end
+
+  def execute(return_domain_in)
```
danielburnley: `def execute(project_id:, data:)` ?

lib/local_authority/use_cases.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class LocalAuthority::UseCases
+  def self.register(builder)
+    builder.define_use_case :return_gateway do
+      LocalAuthority::Gateway::InMemoryTemplate.new
```
danielburnley: :eyes: Was this just to have something there? If so I'd just put `nil`

spec/unit/local_authority/use_case/create_return_spec.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+require 'rspec'
+
+describe LocalAuthority::UseCase::CreateReturn do
```
danielburnley: SUGGEST (see [here](https://github.com/madetech/rfcs/blob/master/CodeReviewStandards/rfc-007-code-review-comments.md)): another place the following might be useful:

```
let(:response) { create_return.execute(return_hash) }
before{ response }
```

This then means your tests don't need to keep calling `create_return.execute` on inside every one, less to refactor if that interface changes

spec/unit/local_authority/use_case/get_return_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+  let(:return_hash) do
+    {
+      project_id: 0,
+      data:  { cats: 'Jerry' }
+    }
+  end
+
+  let(:use_case) { described_class.new(return_gateway: return_gateway).execute(id: 0) }
+
+  before { @response = use_case }
```
danielburnley: Better to use lets rather than instance variables from memory, though I can't think of the reason right now

spec/web_routes/create_return_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+      def post_return
+        post '/return/create',
+             { project_id: 1, data: { cats: 'Meow' } }.to_json, 'Content-Type' => 'application/json'
+        response_body = JSON.parse(last_response.body)
+        response_body['id']
+      end
+    end
+  end
+
+
```
danielburnley: :eyes: whitespace

spec/web_routes/create_return_spec.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+require 'rspec'
+require_relative 'delivery_mechanism_spec_helper'
+
+describe 'Creating returns' do
```
danielburnley: This should be stubbing out the actual use cases and asserting that the correct data is being passed to/from the use cases (see other web route tests for examples)

We already have acceptance tests covering the the use case logic so the web routes tests don't need to test that as well

spec/web_routes/create_return_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+      it 'will return different ids' do
+        id1 = post_return
+        id2 = post_return
+
+        expect(id1).to_not eq(id2)
+      end
+
+      private
+
+      def post_return
```
danielburnley: I don't think this needs to be a private method - but i remember seeing some discussion elsewhere on this so i'm not 100% confident

spec/web_routes/get_return_spec.rb
```diff
@@ -0,0 +1,100 @@
+# frozen_string_literal: true
+
+require 'rspec'
+require_relative 'delivery_mechanism_spec_helper'
+
+describe 'Getting returns' do
```
danielburnley: This also needs to have some stubbing of use cases done

lib/delivery_mechanism/web_routes.rb
```diff
@@ -6,15 +6,44 @@ module DeliveryMechanism
...
       200
     end
 
+    post '/return/create' do
+      body = request.body.read
+      if body.to_s.empty?
+        response.status = 400
+
+      else
+        request_body = Common::DeepSymbolizeKeys.to_symbolized_hash(JSON.parse(body))
```
danielburnley: Having thought about this I think we should avoid just blindly symbolizing the request and throwing it into the use case, it makes it a bit harder to read as to what's actually expected

Putting it as:

```ruby
return_id = @use_case_factory.get_use_case(:create_return).execute(
    project_id: request_body['project_id'],
    data: request_body['data']
)
```

Makes it much clearer what it's actually passing to use cases which makes it a bit easier to see what params are expected in the request

lib/delivery_mechanism/web_routes.rb
```diff
@@ -6,15 +6,44 @@ module DeliveryMechanism
...
+        response.status = 200
+      end
+      response
+    end
+
+    get '/return/get' do
+      return 400 if params['id'].nil?
+
+      return_hash = @use_case_factory.get_use_case(:get_return).execute(id: params['id'].to_i)
+      p "is return hash a nil? #{return_hash.nil?}"
```
danielburnley: :eyes:

lib/delivery_mechanism/web_routes.rb
```diff
@@ -6,15 +6,44 @@ module DeliveryMechanism
...
+      else
+        request_body = Common::DeepSymbolizeKeys.to_symbolized_hash(JSON.parse(body))
+        return_id = @use_case_factory.get_use_case(:create_return).execute(request_body)
+        response.body = { id: return_id[:id]}.to_json
+        response.status = 200
+      end
+      response
+    end
+
+    get '/return/get' do
```
danielburnley: I am so conflicted about these web routes :thinking:

danielburnley: It works for now at least

spec/web_routes/create_return_spec.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+require 'rspec'
+require_relative 'delivery_mechanism_spec_helper'
+
+describe 'Creating returns' do
+  let(:create_return_spy) do Class.new do
```
danielburnley: This could just be a spy then you can control it's return value easily and test against it:

`let(:create_return_spy) { spy(execute: created_return_id) }`

Then in your examples you can do:

`let(:created_return_id) { X }`

Which means your assertions become more accurate than just `to_not eq(nil)` and can be your `created_return_id`

spec/web_routes/create_return_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      end
+
+      it 'will return json with id' do
+        response_body = JSON.parse(last_response.body)
+        expect(response_body['id']).to_not eq(nil)
+      end
+    end
+  end
+
+  context 'with multiple returns' do
```
danielburnley: Do we need to test multiple returns now we stub out the `create_return`?

spec/web_routes/get_return_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+  let(:get_return_spy) do Class.new do
```
danielburnley: This (like the above) could be a spy which you set the return values on, then your tests will be a lot simpler as you just check you're passing the correct arguments to/from the use case (the boundary)

PR #12, opened 2018-07-30
Get Returns

faissalMT:


lib/local_authority/gateway/sequel_return.rb
```diff
@@ -17,5 +17,15 @@ def find_by(id:)
       r.data = Common::DeepSymbolizeKeys.to_symbolized_hash(row[:data].to_h)
     end
   end
-end
 
+  def get_returns(project_id:)
+    returns_array = []
+    @database[:returns].where(project_id: project_id).each do |db_r|
```
danielburnley: you could do:
```
@database[:returns].where(project_id: project_id).all.map do |row|
    LocalAuthority::Domain::Return.new.tap do |r|
        r.pro...
    end
end
```

This saves having to create a reurns_array and push them each manually

lib/local_authority/gateway/sequel_return.rb
```diff
@@ -17,5 +17,15 @@ def find_by(id:)
...
+      returns_array << LocalAuthority::Domain::Return.new.tap do |r|
+        r.project_id = db_r[:project_id]
```
danielburnley: Would be good to also return the ID of the Return here as we'll need it down the line

lib/local_authority/use_case/get_returns.rb
```diff
@@ -0,0 +1,9 @@
+class HomesEngland::UseCase::GetReturns
+  def initialize(return_gateway:)
+    @return_gateway = return_gateway
+  end
+
+  def execute(project_id:)
+    @return_gateway.get_returns(project_id: project_id).map { |r| {project_id: r.project_id, data: r.data}}
```
danielburnley: Would be worth breaking into a few lines for readability, could also move the turning of a return into a hash into a private method in the use case so you can do:

`get_returns(..).map(&:return_to_hash)`

spec/acceptance/local_authority/get_returns_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+          type: 'Cat waiting room',
+          description: 'A waiting room for cats',
+          completion_date: '2019-09-01'
+        },
+        financial: {
+          date: '2019-09-01',
+          funded_through_HIF: true
+        }
+      }))
+    expect(get_use_case(:create_return).execute(project_id: 2, data:
```
danielburnley: shouldn't need the expect around these

spec/unit/local_authority/use_case/get_returns_spec.rb
```diff
@@ -0,0 +1,164 @@
...
+          },
+          financial: {
+            date: '2017-12-25',
+            funded_through_HIF: true
+          }
+        }
+      }]}
+
+    let(:all_return_objects) {[
+      double(
```
danielburnley: I'd use the actual Domain object here rather than a double

spec/unit/local_authority/use_case/get_returns_spec.rb
```diff
@@ -0,0 +1,164 @@
...
+
+describe HomesEngland::UseCase::GetReturns do
+
+  let(:all_returns) { [] }
+  let(:all_return_objects) { [] }
+  let(:return_gateway) { spy(get_returns: all_return_objects)}
+
+  let(:get_returns) {described_class.new(return_gateway: return_gateway)}
+
+  before do
```
danielburnley: you shouldn't need this here as you're dependency injecting the gateway into the use case, not `new`-ing up the gateway

spec/unit/local_authority/use_case/get_returns_spec.rb
```diff
@@ -0,0 +1,164 @@
...
+      stub_const(
+        'HomesEngland::Gateway::ReturnGateway',
+        double(new: return_gateway)
+      )
+  end
+
+  context 'with a single item' do
+    let(:all_returns) {[
```
danielburnley: These returns are pretty long, we don't need them to reflect actual data here just some basic stuff we can test against: 

```
LocalAuthority::Domain::Return.new.tap do |r|
  r.id = 1
  r.project_id = 2
  r.data = { cats: 'meow' }
end
```

would probably be enough

danielburnley: Currently they take up a lot of space and make it not entirely readable

lib/local_authority/use_case/get_returns.rb
faissalMT: I feel like the private method would be overkill

.circleci/config.yml
```diff
@@ -31,13 +32,9 @@ jobs:
       - run:
           name: Waiting for Postgres to be ready
           command: |
-            for i in `seq 1 10`;
-            do
-              nc -z localhost 5432 && echo Success && exit 0
```
danielburnley: sigh

lib/local_authority/gateway/sequel_return.rb
```diff
@@ -17,5 +17,14 @@ def find_by(id:)
       r.data = Common::DeepSymbolizeKeys.to_symbolized_hash(row[:data].to_h)
     end
   end
-end
 
+  def get_returns(project_id:)
+    @database[:returns].where(project_id: project_id).all.map do |db_r|
```
danielburnley: Could call that variable `row` rather than `db_r`, but it's small scoped so not gonna be too picky about it

PR #14, opened 2018-07-30
Save returns in the database

danielburnley:
WHAT - Adds in a sequel gateway for returns

WHY - Whilst writing the usecases we used an in memory gateway for returns, however we need to store these returns in the database


spec/acceptance/local_authority/perform_return_spec.rb
```diff
@@ -6,6 +6,15 @@
 describe 'Performing Return on HIF Project' do
   include_context 'use case factory'
 
+  def create_new_return(return_data)
```
robinlacey: 👍 this makes it more readable

spec/unit/local_authority/gateway/sequel_return_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    it 'creates the new return' do
+      found_return = gateway.find_by(id: return_id)
+
+      expect(found_return.project_id).to eq(1)
+      expect(found_return.data).to eq(cats: "meow")
+    end
+  end
+
+  context 'example one' do
```
robinlacey: example two?

PR #15, opened 2018-07-30
Add get base return use case

robinlacey:


lib/local_authority/domain/return_template.rb
```diff
@@ -0,0 +1,7 @@
+class LocalAuthority::Domain::ReturnTemplate
+  attr_reader :layout
+
+  def initialize(layout:)
+    @layout = layout
```
faissalMT: This is inconsistent with how we're creating domains elsewhere, we rely on tap to assign values.

danielburnley: Yeah you're not wrong - Wondering when you'd use one and not the other but i'm not sure. I'll change it now for consistency :+1:

PR #17, opened 2018-08-01
Schema for project

danielburnley:


lib/homes_england/gateway/in_memory_template.rb
```diff
@@ -5,25 +5,77 @@ def initialize; end
 
   def get_template(type:)
     return nil unless type == 'hif'
-    HomesEngland::Domain::Template.new(
-      layout: {
-        summary: {
-          project_name: nil,
-          description: nil,
-          lead_authority: nil
+    hif_template = HomesEngland::Domain::Template.new
```
robinlacey: I know we're using tap elsewhere - which i think would be ugly here - but i think it might be a bit nicer to pass in the `data` and `schema` in Template initialize

spec/unit/homes_england/use_case/create_new_project_spec.rb
```diff
@@ -21,17 +19,11 @@
...
-    it 'populate template with baseline data receives type baseline data ' do
-      expect(populate_template_use_case).to have_received(:execute)
-        .with(type: 'hif', baseline: baseline)
-    end
 
     it 'creates the project with populated data' do
       expect(project_gateway).to have_received(:create) do |project|
         expect(project.type).to eq('hif')
-        expect(project.data).to eq(key: 'newValue')
+        expect(project.data).to eq(key: 'value')
```
robinlacey: assume we're not using the project.schema because the type will fetch the latest?

lib/homes_england/gateway/in_memory_template.rb
faissalMT: I'm ok with this so long as we switch other uses of tap for domains.

PR #18, opened 2018-08-02
GetReturns can now be accessed via /project/:id/returns

faissalMT:
Also addresses some mild errors in my last pull

### DIFF
--- a/lib/delivery_mechanism/web_routes.rb
+++ b/lib/delivery_mechanism/web_routes.rb
@@ -63,6 +63,12 @@ module DeliveryMechanism
       end
     end
 
+    get '/project/:id/returns' do
+      returns = @use_case_factory.get_use_case(:get_returns).execute(project_id: params['id'].to_i)
+      response.status = returns.empty? ? 404 : 200
+      response.body = returns.to_json
+    end
+
     get '/project/find' do
       return 404 if params['id'].nil?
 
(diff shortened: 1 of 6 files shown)


spec/web_routes/get_returns_spec.rb
```diff
@@ -0,0 +1,83 @@
+# frozen_string_literal: true
+
+require 'rspec'
+require_relative 'delivery_mechanism_spec_helper'
+
+describe 'Getting return history' do
```
danielburnley: These tests look really good, only thing I'd add is just a test for each example to make sure we're calling the `get_returns_spy` with the correct ID

lib/delivery_mechanism/web_routes.rb
```diff
@@ -63,6 +63,12 @@ class WebRoutes < Sinatra::Base
       end
     end
 
+    get '/project/:id/returns' do
+      returns = @use_case_factory.get_use_case(:get_returns).execute(project_id: params['id'].to_i)
+      response.status = returns.empty? ? 404 : 200
+      response.body = returns.to_json
```
danielburnley: I'd put the array in a key: `{returns: returns}.to_json`

Returning just an array over JSON feels... off? I'm not sure on convention but it feels better to give it a name

PR #20, opened 2018-08-02
Update return

robinlacey:
**What**
When a return is Created it is set the status of 'Draft'
Returns can be hard updated via `post` `return/create` 

**Why**
HE need to have draft returns before they're ready for Review

### DIFF
--- a/lib/local_authority/use_case/create_return.rb
+++ b/lib/local_authority/use_case/create_return.rb
@@ -3,10 +3,11 @@ class LocalAuthority::UseCase::CreateReturn
     @return_gateway = return_gateway
   end
 
-  def execute(project_id:, data:)
+  def execute(project_id:, data:, status: 'Draft')
     return_object = LocalAuthority::Domain::Return.new.tap do |r|
       r.project_id = project_id
       r.data = data
+      r.status = status
     end
 
     { id: @return_gateway.create(return_object) }
--- a/lib/local_authority/use_case/create_return.rb
+++ b/lib/local_authority/use_case/create_return.rb
@@ -3,11 +3,10 @@ class LocalAuthority::UseCase::CreateReturn
     @return_gateway = return_gateway
   end
 
-  def execute(project_id:, data:, status: 'Draft')
+  def execute(project_id:, data:)
     return_object = LocalAuthority::Domain::Return.new.tap do |r|
       r.project_id = project_id
       r.data = data
-      r.status = status
     end
 
     { id: @return_gateway.create(return_object) }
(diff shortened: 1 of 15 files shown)


lib/local_authority/use_case/create_return.rb
```diff
@@ -3,10 +3,11 @@ def initialize(return_gateway:)
     @return_gateway = return_gateway
   end
 
-  def execute(project_id:, data:)
+  def execute(project_id:, data:, status: 'Draft')
```
danielburnley: :eyes: Feels weird to have a default of 'draft' - given `create_return` will always be creating a draft, I'd hard code it in the creation of the domain object (line 10)

lib/local_authority/use_case/submit_return.rb
```diff
@@ -0,0 +1,9 @@
+class LocalAuthority::UseCase::SubmitReturn
+  def initialize(return_gateway:)
+    @return_gateway = return_gateway
+  end
+
+  def execute(return_id:)
+    @return_gateway.submit(return_id: return_id)
```
danielburnley: Picky - but I'd clearly show what this use case returns rather than returning stuff directly from the gateway - but I'm not going to say it has to be done for this PR

lib/local_authority/use_case/update_return.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class LocalAuthority::UseCase::UpdateReturn
+  def initialize(return_gateway:)
+    @return_gateway = return_gateway
+  end
+
+  def execute(return_id:, data:)
+    @return_gateway.update(return_id: return_id,data: data)
```
danielburnley: :thinking: I'd have this return success true/false potentially, not sure how important it is right now though but i'm wary of how many "we can do this later" tasks we're creating

lib/local_authority/use_case/create_return.rb
```diff
@@ -3,10 +3,11 @@ def initialize(return_gateway:)
     @return_gateway = return_gateway
   end
 
-  def execute(project_id:, data:)
```
danielburnley: I'm in two minds on whether or not create_new_return should actually take data anymore.

Imo it should be:

def execute(project_id:)

and create a new return with the baseline data from the project

Otherwise we need to have some way of creating a new project without knowing the initial data... i don't think

lib/local_authority/use_case/submit_return.rb
faissalMT: This use case's return isn't important, we've considered just returning nil but that doesn't seem very idiomatic

PR #33, opened 2018-08-13
Specify baseline key in return schema to populate with data

faissaloo:
Note that this will break when merged because of changes to the in memory template are needed.  
Cleanup is also still needed


spec/unit/local_authority/use_case/find_baseline_path_spec.rb
```diff
@@ -0,0 +1,143 @@
+#Yes I know this shouldn't be a use case but icba to move it at this point
```
robinlacey: 😢

PR #34, opened 2018-08-13
 Specify baseline key in return schema to populate with data

faissalMT:
Why - So returns for future contracts can be created easily, and the return for the current hif contract can contain the place it derives its data from so that it is easily changeable.   
What - A system that allows return schemas to contain paths that point to data in a baseline to fill in data in the return.


lib/local_authority/use_case/populate_return_template.rb
```diff
@@ -1,31 +1,65 @@
...
+    { populated_data: populated_return }
   end
 
   private
+  def bury_hash(hash, path, path_types, value)
+    if path.empty?
+      value
+    else
+      if path_types.first == :array
+        hash[path.first] = [] if hash[path.first].nil?
```
robinlacey: could split this into two functions for readability

lib/local_authority/use_case/populate_return_template.rb
```diff
@@ -1,31 +1,65 @@
...
+      end
+      hash
+    end
+  end
 
-  def populate_hash(base, data)
-    base.each do |key, value|
-      next if data[key].nil?
-      base[key] = get_data_value(base, data, key, value)
+  def bury_array(array, path, path_types, value)
```
robinlacey: perhaps change the name to something a bit longer a descriptive? Not sure what 'bury' means

lib/local_authority/use_case/populate_return_template.rb
```diff
@@ -1,31 +1,65 @@
...
-    populated_data = populate_hash(template.layout.dup, data)
-    {
-      populated_data: populated_data
-    }
+  def execute(type:, baseline_data:)
+    populated_return = {}
+    template_schema = @template_gateway.find_by(type: type).schema
+    @get_schema_copy_paths.execute(type: type)[:paths].each do |copy_paths|
+      path_types = schemaTypes(template_schema, copy_paths[:to]).drop(1)
+      bury_hash(
```
robinlacey: as we're passing a lot to `bury_hash` we could move some of this to a longer/more descriptive method?

lib/local_authority/use_case/populate_return_template.rb
```diff
@@ -1,31 +1,98 @@
...
-    template = @template_gateway.find_by(type: type)
-    populated_data = populate_hash(template.layout.dup, data)
-    {
-      populated_data: populated_data
-    }
+  def execute(type:, baseline_data:)
+    populated_return = {}
+    template_schema = @template_gateway.find_by(type: type).schema
+    @get_schema_copy_paths.execute(type: type)[:paths].each do |copy_paths|
+      path_types = schemaTypes(template_schema, copy_paths[:to]).drop(1)
```
danielburnley: `schemaTypes` -> `schema_types`

spec/unit/local_authority/use_case/find_baseline_path_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+      let(:path) { [:sounds, :dog] }
+
+      it 'can find a relevent item for a single item array' do
+        expect(use_case).to eq(found: ['woof'])
+      end
+    end
+  end
+
+  context 'with paths that include a multi item array' do
+    context 'example 1' do
```
danielburnley: where you've not got example 2 you could either remove the `example` context, or add a second example. I think the code by this point will not have been slimed so you might not have required extra examples

danielburnley: The flipside being that you might want to ensure in the future with more robust examples

PR #35, opened 2018-08-16
Generate token (WIP)

robinlacey:
WHAT 

To view a page a user requires an `access token`. When page is visited with correct `access token` it will put an `API key` in the user header 

An `access token` can be generated via the api `/request/token` end point. It will be emailed (via gov notify api) to a whitelisted email address only.  The email will have a link to the requested page with the `access token` set in the param

WHY 

LA Monitoring tool requires a simple email based authentication.


lib/delivery_mechanism/web_routes.rb
```diff
@@ -19,6 +19,30 @@ class WebRoutes < Sinatra::Base
...
+        email_address: request_hash[:email_address]
+      )[:valid]
+        @use_case_factory.get_use_case(:send_notification).execute(
+          to: request_hash[:email_address], url: request_hash[:url]
+        )
+      end
+      200
+    end
+
+    post '/token/expend' do
```
mrosel: This doesn't appear to be RESTful
can we make this a `DELETE` method?

lib/local_authority/gateway/gov_email_notification_gateway.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'notifications/client'
+class LocalAuthority::Gateway::GovEmailNotificationGateway
+  def initialize
+    @client = Notifications::Client.new(ENV.fetch('GOV_NOTIFY_API_KEY'))
+    @default_template = 'b8fc89b6-79c6-491f-9872-60e110130e4a'
```
mrosel: is this meant to be a secret?

lib/local_authority/use_case/check_api_key.rb
```diff
@@ -0,0 +1,9 @@
+class LocalAuthority::UseCase::CheckApiKey
+  def initialize(api_key_gateway:)
+    @api_key_gateway = api_key_gateway
+  end
+
+  def execute(api_key:)
+    { valid: !@api_key_gateway.find_by(api_key: api_key).nil? }
```
mrosel: Might be an easier way to read/write this
By convention, `find_by` methods should return nil or the found object - additionally, ruby validates `nil` as `false', so this could be shortened to
`@api_key_gateway.find_by(api_key: api_key)`

lib/local_authority/use_case/expend_access_token.rb
```diff
@@ -0,0 +1,16 @@
+class LocalAuthority::UseCase::ExpendAccessToken
+  def initialize(access_token_gateway:)
+    @access_token_gateway = access_token_gateway
+  end
+
+  def execute(access_token:)
+    exists = !@access_token_gateway.find_by(access_token: access_token).nil?
```
mrosel: you don't need this variable
` if @access_token_gateway.find_by(access_token: access_token)`
will behave the same way

mrosel: wouldn't it be better to have access_token be created and call the methods on itself?

```ruby
if @access_token
 @access_token.destroy_me
```

lib/local_authority/use_case/create_access_token.rb
```diff
@@ -0,0 +1,15 @@
...
+require 'securerandom'
+class LocalAuthority::UseCase::CreateAccessToken
+
+  def initialize(access_token_gateway:)
+    @access_token_gateway = access_token_gateway
+  end
+
+  def execute
+    access_token = SecureRandom.uuid
+    @access_token_gateway.save(access_token: access_token)
```
danielburnley: Use cases should always communicate with gateways via domain objects ;)

This way if we changed from UUIDs to something else, the actual gateway wouldn't need to change as long as the domain object had the same attributes

lib/local_authority/use_case/create_access_token.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require 'securerandom'
+class LocalAuthority::UseCase::CreateAccessToken
```
danielburnley: Are these LocalAuthority specific, or should this belong in common as they have common actors

lib/local_authority/use_case/expend_access_token.rb
danielburnley: In Clean Architecture the access token would be a domain object, which would be passed to a gateway for deletion. The way you described would mean the domain object would have knowledge of the database and have multiple reasons to change and violate the single responsibility principle

lib/local_authority/gateway/gov_email_notification_gateway.rb
robinlacey: the template? not really, we thought i'd be overkill to make a env var :)

lib/local_authority/use_case/create_access_token.rb
faissalMT: They belong in common and will be moved before merging.

lib/delivery_mechanism/web_routes.rb
robinlacey: the API isn't currently RESTful (hence `project/create` and `project/update`) - we're aiming to show intent in naming over methods. I'm a fan of RESTful but others .. not so much 👀 @danielburnley

lib/local_authority/use_case/send_notification.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
```
StevenLeighton21: What's the reasoning for freezing strings in this file?

spec/web_routes/request_access_token_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    expect(check_email_spy).to have_received(:execute).with(email_address: valid_email)
+  end
+
+  context 'given a valid email address' do
+    it 'returns a 200' do
+      expect(last_response.status).to eq(200)
+    end
+
+    it 'passes email address and url to send notification use case' do
+      expect(send_notification_spy).to have_received(:execute).with(to: valid_email, url: 'http://catscatscats.cat')
```
StevenLeighton21: good URL

lib/local_authority/use_case/send_notification.rb
robinlacey: it's a rubocop thang : https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/FrozenStringLiteralComment

spec/web_routes/request_access_token_spec.rb
robinlacey: 🐱

lib/delivery_mechanism/web_routes.rb
```diff
@@ -31,18 +57,30 @@ class WebRoutes < Sinatra::Base
...
-      return 400 if request_hash.nil?
-
-      return_id = @use_case_factory.get_use_case(:create_return).execute(
-        project_id: request_hash[:project_id],
-        data: request_hash[:data]
-      )
+    def guard_access(env, request)
+      if env['HTTP_API_KEY'].nil?
+        response.status = 400
+      elsif !@use_case_factory.get_use_case(:check_api_key).execute(api_key: env['HTTP_API_KEY'])[:valid]
```
mrosel: `env`'s are everywhere!

lib/delivery_mechanism/web_routes.rb
```diff
@@ -31,18 +57,30 @@ class WebRoutes < Sinatra::Base
...
+      end
+    end
 
-      response.tap do |r|
-        r.body = { id: return_id[:id] }.to_json
-        r.status = 201
+    post '/return/create' do
+      guard_access env, request do |request|
+        request_hash = get_hash(request)
+        return 400 if request_hash.nil?
```
mrosel: You definitely don't need to change this, but I always felt this is the only time unless is useful and reads better...
`return 400 unless request_hash`

lib/local_authority/use_case/send_notification.rb
StevenLeighton21: I was interested in the reasoning for doing so, if it's performance that's cool (I didn't think about that). Neat!

PR #36, opened 2018-08-20
/token/request calls create token usecase

robinlacey:
send notifications are sent token so url can be created correctly

### DIFF
--- a/lib/local_authority/gateway/gov_email_notification_gateway.rb
+++ b/lib/local_authority/gateway/gov_email_notification_gateway.rb
@@ -7,7 +7,11 @@ class LocalAuthority::Gateway::GovEmailNotificationGateway
     @default_template = 'b8fc89b6-79c6-491f-9872-60e110130e4a'
   end
 
-  def send_notification(to:, url:)
-    @client.send_email(email_address: to, template_id: @default_template, personalisation: { access_url: url })
+  def send_notification(to:, url:, access_token:)
+    @client.send_email(
+      email_address: to,
+      template_id: @default_template,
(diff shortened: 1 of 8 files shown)


lib/local_authority/gateway/gov_email_notification_gateway.rb
```diff
@@ -7,7 +7,11 @@ def initialize
     @default_template = 'b8fc89b6-79c6-491f-9872-60e110130e4a'
   end
 
-  def send_notification(to:, url:)
-    @client.send_email(email_address: to, template_id: @default_template, personalisation: { access_url: url })
+  def send_notification(to:, url:, access_token:)
+    @client.send_email(
+      email_address: to,
+      template_id: @default_template,
+      personalisation: { access_url: url + "/?token=#{access_token}" }
```
chrisblackburn: Would consider not mixing concatenation and interpolation

robinlacey: 😪

lib/local_authority/use_case/send_notification.rb
```diff
@@ -5,7 +5,7 @@ def initialize(notification_gateway:)
     @notification_gateway = notification_gateway
   end
 
-  def execute(to:, url:)
-    @notification_gateway.send_notification(to: to, url: url)
+  def execute(to:, url:, access_token:)
+    @notification_gateway.send_notification(to: to, url: url,access_token: access_token)
```
chrisblackburn: Missing a space

PR #44, opened 2018-08-23
Returns from returns

faissalMT:
What - Return schemas can now reference a path in the previous returns. The schema path format has changed slightly, it must now be prefixed with the data source for example `[:baseline_data, :cats]` vs `[:return_data, :dogs]` the schema key used to store this path has also changed from `:baselineKey` to `:sourceKey`. Note that there is not currently any way of referencing a specific return.  
  
Why - This is so that returns using relevant and up to date data can be created.

### DIFF
--- a/lib/local_authority/gateway/in_memory_return_template.rb
+++ b/lib/local_authority/gateway/in_memory_return_template.rb
@@ -1299,7 +1299,8 @@ class LocalAuthority::Gateway::InMemoryReturnTemplate
                                   hifSpendLastReturn: {
                                     type: 'integer',
                                     readonly: true,
-                                    title: 'TO BE CALCULATED - Current HIF spend since last return'
+                                    title: 'TO BE CALCULATED - Current HIF spend since last return',
+                                    sourceKey: [:return_data, :funding, :fundingPackages, :fundingPackage, :overview, :hifSpendSinceLastReturn, :hifSpendCurrentReturn]
                                   },
                                   hifSpendVariance: {
                                     type: 'integer',
--- a/lib/local_authority/gateway/in_memory_return_template.rb
+++ b/lib/local_authority/gateway/in_memory_return_template.rb
@@ -21,12 +21,12 @@ class LocalAuthority::Gateway::InMemoryReturnTemplate
                   type: 'string',
                   # This is marked as a dropdown but also from baseline?
                   title: 'Type',
-                  baselineKey: [:infrastructures, :type]
+                  sourceKey: [:baseline_data, :infrastructures, :type]
                 },
                 description: {
                   type: 'string',
                   title: 'Description',
-                  baselineKey: [:infrastructures, :description]
+                  sourceKey: [:baseline_data, :infrastructures, :description]
                 },
                 planning: {
                   type: 'object',
@@ -37,7 +37,7 @@ class LocalAuthority::Gateway::InMemoryReturnTemplate
                     baselineOutlinePlanningPermissionGranted: {
                       type: 'boolean',
                       title: 'Outline Planning Permission granted',
-                      baselineKey: [:infrastructures, :outlinePlanningStatus, :granted]
+                      sourceKey: [:baseline_data, :infrastructures, :outlinePlanningStatus, :granted]
                     },
                     planningNotGranted: {
                       type: 'object',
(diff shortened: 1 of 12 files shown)


lib/local_authority/use_case/get_base_return.rb
```diff
@@ -1,17 +1,24 @@
...
-  def initialize(return_gateway:, project_gateway:, populate_return_template:)
+  def initialize(return_gateway:, project_gateway:, populate_return_template:, get_returns:)
     @return_gateway = return_gateway
     @project_gateway = project_gateway
     @populate_return_template = populate_return_template
+    @get_returns = get_returns
   end
 
   def execute(project_id:)
+    return_data = @get_returns.execute(project_id: project_id).dig(:returns, -1, :updates, -1)
```
danielburnley: I'd move `@get_returns.execute(project_id: project_id).dig(:returns, -1, :updates, -1)` out into a function called `get_last_return_for_project(project_id)`, to help better express what it's doing as reading the dig doesn't really help me

lib/local_authority/use_case/get_base_return.rb
```diff
@@ -1,17 +1,24 @@
...
     project = @project_gateway.find_by(id: project_id)
     schema = @return_gateway.find_by(type: project.type)
-    data = @populate_return_template.execute(type: project.type,
-      baseline_data: project.data)[:populated_data]
+    if return_data.nil?
```
danielburnley: move this out to a function maybe? `populate_return(project_data, return_data)` Helps understand what that particular `if` block is doing

lib/local_authority/use_case/populate_return_template.rb
```diff
@@ -5,16 +5,19 @@ def initialize(template_gateway:, find_baseline_path:, get_schema_copy_paths:)
...
   end
 
-  def execute(type:, baseline_data:)
+  def execute(type:, baseline_data:, return_data: {})
+    source_data = { baseline_data: baseline_data, return_data: return_data }
     populated_return = {}
     template_schema = @template_gateway.find_by(type: type).schema
-    @get_schema_copy_paths.execute(type: type)[:paths].each do |copy_paths|
+    paths = @get_schema_copy_paths.execute(type: type)[:paths]
+    paths.each do |copy_paths|
```
danielburnley: This block would also be good to move into a function describing what it's doing, as a person maintaining it I would like to know what it's doing

spec/acceptance/local_authority/perform_return_spec.rb
```diff
@@ -396,6 +396,53 @@ def expect_return_to_be_submitted(id:)
 
     submit_return(id: return_id)
     expect_return_to_be_submitted(id: return_id)
+
+    second_return = {
```
danielburnley: As this acceptance spec is from the perspective of a user, I would expect there to be an assertion to show that data is being pulled from prior returns, similar to how the test starts with a call to get_base_return

danielburnley: As a user - I expect to get the base return and see it populated from the previous submitted return

lib/local_authority/use_case/get_base_return.rb
danielburnley: What if the last return was a draft which was abandoned? Might want to only get the last submitted return

spec/unit/local_authority/use_case/get_base_return_spec.rb
```diff
@@ -1,98 +1,332 @@
 
+#We now need to get the last update and factor it into our populate
```
danielburnley: :eyes:

spec/acceptance/local_authority/perform_return_spec.rb
```diff
@@ -295,6 +295,136 @@ def expect_return_to_be_submitted(id:)
     }
   end
 
+  let(:expected_second_base_return) do
```
robinlacey: we need to come up with a better way of doing this - lets chat about off line

lib/local_authority/gateway/in_memory_return_template.rb
```diff
@@ -1299,7 +1299,8 @@ def find_by(type:)
                                   hifSpendLastReturn: {
                                     type: 'integer',
                                     readonly: true,
-                                    title: 'TO BE CALCULATED - Current HIF spend since last return'
+                                    title: 'TO BE CALCULATED - Current HIF spend since last return',
+                                    sourceKey: [:return_data, :funding, :fundingPackages, :fundingPackage, :overview, :hifSpendSinceLastReturn, :hifSpendCurrentReturn]
```
robinlacey: not for this PR, but it'd be nice to have this something along the lines of `previous_submitted_data` ... but not terrible

PR #48, opened 2018-08-28
Use database and store project ids

faissalMT:
Also includes acceptance tests for unimplemented per-project authentication.

### DIFF
--- a/lib/homes_england/use_case/add_user.rb
+++ b/lib/homes_england/use_case/add_user.rb
@@ -4,6 +4,10 @@ class HomesEngland::UseCase::AddUser
   end
 
   def execute(email:)
-    @user_gateway.create(email)
+    @user_gateway.create(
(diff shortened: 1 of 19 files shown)


lib/homes_england/use_case/add_user.rb
```diff
@@ -0,0 +1,13 @@
+class HomesEngland::UseCase::AddUser
+  def initialize(user_gateway:)
+    @user_gateway = user_gateway
+  end
+
+  def execute(email:)
+    @user_gateway.create(
+      LocalAuthority::Domain::User.new.tap do |u|
```
danielburnley: To make this a bit nicer i'd assign the user to a variable then pass it in, just looks a bit cleaner than nested stuff imo

danielburnley: Not required though, just something for the future

PR #49, opened 2018-08-29
User, access token and API key permissions

faissalMT:
What - Restricts user, access token and api key access to only specified projects


jpmaHub:
It advised to write your commit messages in the imperative, so that you’re following the Git’s own built-in convention. Also, It would be better to squash or rebase your commits because some of the commits are failing.

spec/acceptance/local_authority/authorise_user_spec.rb
```diff
@@ -3,7 +3,7 @@
 require 'rspec'
 require_relative '../shared_context/use_case_factory'
 
-describe 'Authorises the user' do
+xdescribe 'Authorises the user' do
```
robinlacey: assume this is still wip?

faissalMT: No this needs to be removed

faissalMT: No it needs to be removed.

PR #50, opened 2018-08-31
Add tests for ensuring returns have ids

faissalMT:
Came across a bug while writing my calculations acceptance spec where ids were nil.


spec/unit/local_authority/gateway/sequel_return_spec.rb
```diff
@@ -23,6 +23,11 @@
       expect(found_return.project_id).to eq(1)
       expect(found_return.status).to eq('Draft')
     end
+
+    it 'gives the returns ids' do
+      found_return = gateway.find_by(id: return_id)
+      expect(found_return.id).not_to be_nil
```
robinlacey: do we want to check what this is equal to?

faissalMT: It doesn't particularly matter what it's equal to as long as it's unique so I guess we could do a test for uniqueness if you really want to.

robinlacey: might be a good idea, this could be slimed?

PR #51, opened 2018-08-31
Return calculations

faissalMT:
What - Calculate certain fields based on provided data


lib/local_authority/use_case/calculate_hif_return.rb
```diff
@@ -0,0 +1,70 @@
...
+        date_as_string_one: get_currentReturn(return_data_with_no_calculations),
+        date_as_string_two: get_currentReturn(previous_return)
+      )
+    )
+    {
+      calculated_return: new_return_data
+    }
+  end
+
+  def get_currentReturn(return_data)
```
danielburnley: `private` ?

lib/local_authority/use_case/calculate_hif_return.rb
```diff
@@ -0,0 +1,70 @@
...
+            }
+          }
+        }
+      }
+    }
+
+    new_return_data = deep_merge(return_data_with_no_calculations, expected_return_data)
+    update_varianceLastReturnFullPlanningPermissionSubmitted(
+      new_return_data,
+      get_currentReturn(previous_return).nil? ? nil : week_difference(
```
danielburnley: I'd move this to a method because multiline ternary statements are not the best to read

```ruby
def foo(*)
    return nil if get_currentReturn(previous_return).nil?
    <the rest of the stuff>
end
```

Reads much nicer than having to read a ternary

spec/unit/local_authority/use_case/calculate_hif_return_spec.rb
```diff
@@ -0,0 +1,232 @@
...
+          }
+        }
+
+        return_data_with_no_calculation = {
+          infrastructures: {
+            planning: {
+              planningNotGranted: {
+                fieldOne: {
+                  returnInput: {
+                    CurrentReturn: '03/11/1982'
```
danielburnley: Shouldn't this be `currentReturn` with a lowercase? 

https://github.com/homes-england/monitor-api/blob/master/lib/local_authority/gateway/in_memory_return_template.rb#L145

lib/local_authority/use_case/calculate_hif_return.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def week_difference(date_as_string_one:, date_as_string_two:)
+    date_one = Date.parse date_as_string_one
+    date_two = Date.parse date_as_string_two
+
+    (date_one.cweek - date_two.cweek).to_s
+  end
+
+  # based on : https://stackoverflow.com/a/30225093
+  def deep_merge(first, second)
```
danielburnley: Rather than this `deep_merge`, maybe creating a helper method which creates the paths as you go might be the better solution?

e.g. `hash.assign_path([:a, :b, :c, :d], value)`

This could loop over the keys and where it's nil assign it to an empty object, then set the final one to the value

Not sure which is the better solution though, this method just scares me :|

faissalMT: I feel like that's less straightforward than this solution because with this I can actually see the data structure I'm injecting.

lib/local_authority/use_case/get_return.rb
```diff
@@ -1,15 +1,29 @@
...
+    project_id = found_return.project_id
     updates = @return_update_gateway.updates_for(return_id: id)
 
+
+    previous_return = @get_returns.execute(project_id: project_id)[:returns].select do |return_data|
+      return_data[:id] < found_return.id && return_data[:status] == 'Submitted'
+    end
+
+    previous_return_data = previous_return.dig(-1, :updates, -1)
+    # The acceptance spec doesn't have :data that's why it's not working, find out which one is correct
```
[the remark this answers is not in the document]
faissalMT: 🤦‍♂️

lib/local_authority/use_case/calculate_hif_return.rb
danielburnley: Yeah that's true - Probably lean towards this solution then

PR #52, opened 2018-09-03
Pass project_ids to check_email use case in webroutes

faissalMT:
What - Fixes authentication per project via the webroutes

### DIFF
--- a/lib/delivery_mechanism/web_routes.rb
+++ b/lib/delivery_mechanism/web_routes.rb
@@ -58,19 +58,8 @@ module DeliveryMechanism
       200
     end
 
-    def guard_access(env, request)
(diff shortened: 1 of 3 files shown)


lib/delivery_mechanism/web_routes.rb
```diff
@@ -201,6 +190,53 @@ def get_hash(request)
...
+      if access_status == :bad_request
+        response.status = 400
+      elsif access_status == :forbidden
+        response.status = 401
+      else
+        yield request_hash
+      end
+    end
+
+    def get_access_status(env, params, request_hash)
```
danielburnley: :man_facepalming: 

We need to make our web stuff consistent as soon as we have some time

PR #54, opened 2018-09-05
Add project/:id endpoint

goingdotin:
This PR adds project/:id endpoint that will either return project data in its response body or 401 depending on whether api key is valid given project id


danielburnley:
Be interested in what value this adds over our current existing endpoint? 

We've got a lot of inconsistencies at the moment anyway, so I think as we touch specific endpoints its worth making things better, but duplicating behaviour is going to cause issues 

If we're doing this for the sake of consistency, then I'd suggest: 

- Extract the behaviour for the endpoint to a controller
- Register this endpoint with that controller
- Register the other endpoint with that controller 

As long the tests still pass we're good to: 

- update the front end to use the new endpoint
- subsequently delete the old endpoint

This way we can avoid duplicating behaviour in the meantime and also avoid breaking our frontend

goingdotin:
I did not realise we already have an endpoint for doing this, if that's there is no reason to reinvent the wheel tbh

danielburnley:
Admittedly `/project/find` appears not to be behind the `guard_access` function so it'd be worth adding that in as a PR anyway, which would be a Good Thing

I'll expand this with some examples when I get in tomorrow morning

danielburnley:
[Here](https://github.com/homes-england/monitor-api/blob/master/lib/delivery_mechanism/web_routes.rb#L140) is the existing Project Find, and [here](https://github.com/homes-england/monitor-api/blob/master/lib/delivery_mechanism/web_routes.rb#L118) is an example of our current `guard_access` method

danielburnley:
Though any updates to these endoints will need some updating on the frontend too, but that's easily done

PR #55, opened 2018-09-06
Add /project/:id/add_users route

goingdotin:
This PR adds a post route to add multiple users to a project. Such a route is protected behind admin api key that should be hidden within environment variables

### DIFF
--- a/lib/delivery_mechanism/web_routes.rb
+++ b/lib/delivery_mechanism/web_routes.rb
@@ -172,6 +172,22 @@ module DeliveryMechanism
       response.status = 201
     end
 
+    post '/project/:id/add_users' do
+      guard_admin_access env, params, request do |request_hash|
+        user_emails = request_hash[:users]
+
+        return 400 unless user_emails.instance_of? Array
+        cleaned_up_emails = user_emails.map(&:strip).reject(&:empty?)
+        return 400 if cleaned_up_emails.empty?
+
+        use_case = @use_case_factory.get_use_case(:add_user_to_project)
+        cleaned_up_emails.each do |email|
+          use_case.execute({ email: email, project_id: params[:id].to_i })
+        end
+        200
+      end
+    end
+
     post '/project/update' do
       request_hash = get_hash(request)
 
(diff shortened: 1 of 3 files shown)


lib/delivery_mechanism/web_routes.rb
```diff
@@ -190,6 +206,29 @@ def get_hash(request)
       end
     end
 
+    def guard_admin_access(env, params, request)
```
danielburnley: The API key should be in the header rather than the request so it'll be `env['HTTP_API_KEY']`

danielburnley: This method seems quite complex for just checking a header also, you could simply have a method:

```ruby
def valid_admin_api_key?(env)
    return false unless env['HTTP_API_KEY'] == ENV['ADMIN_API_KEY']
    true
end
```

Then in the endpoint, rather than doing a guard -> block thing, we can just do

```ruby
return 401 unless valid_admin_api_key?(env)
<rest of the method>
```

spec/web_routes/add_users_to_project_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+require 'rspec'
+require_relative 'delivery_mechanism_spec_helper'
+
+describe 'Adding users to a project' do
+  let(:valid_admin_api_key) { 'supersecret' }
+
+  before do
+    ENV['ADMIN_HTTP_API_KEY'] = valid_admin_api_key
+  end
+  # context 'when incorrect admin http key provided' do
```
danielburnley: Should avoid commiting commented out code

lib/delivery_mechanism/web_routes.rb
```diff
@@ -172,6 +172,22 @@ def get_hash(request)
       response.status = 201
     end
 
+    post '/project/:id/add_users' do
+      guard_admin_access env, params, request do |request_hash|
+        user_emails = request_hash[:users]
+
+        return 400 unless user_emails.instance_of? Array
+        cleaned_up_emails = user_emails.map(&:strip).reject(&:empty?)
+        return 400 if cleaned_up_emails.empty?
```
danielburnley: I'd argue an empty array is a valid request, it just doesn't do anything

lib/delivery_mechanism/web_routes.rb
```diff
@@ -172,6 +172,22 @@ def get_hash(request)
       response.status = 201
     end
 
+    post '/project/:id/add_users' do
+      guard_admin_access env, params, request do |request_hash|
+        user_emails = request_hash[:users]
```
danielburnley: I'd like to start actually using controllers - So it'd be good to move this logic into a controller (see: https://github.com/homes-england/monitor-api/pull/55/files#diff-0da88af6463dac1370811b7c7765fa2dR25)

```ruby
class PostAddUsersToProject
    def execute(request_hash, response)
        ...
        response
    end
end
```

danielburnley: Reason for this being that this file is already quite messy and we need to start being proactive and not add to this mess

lib/delivery_mechanism/web_routes.rb
goingdotin: Done

goingdotin: Done

lib/delivery_mechanism/web_routes.rb
goingdotin: Done

lib/delivery_mechanism/web_routes.rb
goingdotin: Done

PR #57, opened 2018-09-06
Refactor endpoint, guard it from non admins

goingdotin:
This PR refactors code in /project/create and wraps around guarding block that will prevent people without possession of admin api key from using the endpoint. 

It also refactors some of the test setups.

### DIFF
--- a/spec/admin_context.rb
+++ b/spec/admin_context.rb
@@ -1,0 +1,9 @@
+RSpec.shared_context 'as admin' do
+  def set_correct_auth_header
+    header 'API_KEY', 'supersecret'
+  end
+
+  def set_incorrect_auth_header
+    header 'API_KEY', 'wrongsecret'
+  end
+end
--- a/spec/admin_context.rb
+++ b/spec/admin_context.rb
@@ -1,9 +1,0 @@
-RSpec.shared_context 'as admin' do
-  def set_correct_auth_header
-    header 'API_KEY', 'supersecret'
-  end
-
-  def set_incorrect_auth_header
-    header 'API_KEY', 'wrongsecret'
-  end
-end
--- a/spec/web_routes/create_new_project_spec.rb
+++ b/spec/web_routes/create_new_project_spec.rb
@@ -7,9 +7,10 @@ describe 'Creating a new project' do
   include_context 'as admin'
 
   let(:create_new_project_spy) { spy(execute: project_id) }
+  let(:setup_auth_headers) { set_correct_auth_header }
 
   before do
-    set_correct_auth_header
+    setup_auth_headers
 
     stub_const(
       'HomesEngland::UseCase::CreateNewProject',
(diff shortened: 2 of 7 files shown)


spec/web_routes/create_new_project_spec.rb
```diff
@@ -5,8 +5,17 @@
...
 describe 'Creating a new project' do
   let(:create_new_project_spy) { spy(execute: project_id) }
+  let(:valid_admin_api_key) { 'supersecret' }
+
+  def set_correct_auth_header
+    header 'API_KEY', 'supersecret'
+  end
 
   before do
+    ENV['ADMIN_HTTP_API_KEY'] = valid_admin_api_key
```
danielburnley: Where's your failing test you wrote prior to putting the guard in place ;) 

You should put the test in though, to make sure it's 401 on non-auth'd requests

danielburnley:
Also need to put a basic token as an environment variable inside `docker-compose.yml`, anything will do

spec/admin_context.rb
```diff
@@ -0,0 +1,9 @@
+RSpec.shared_context 'as admin' do
+  def set_correct_auth_header
+    header 'API_KEY', 'supersecret'
```
danielburnley: I'd set this to `ENV['ADMIN_HTTP_API_KEY']` rather than to `'supersecret'` - as this won't work in CI unless we set the env variable there too.

Typically your tests should do all the setup required to run in an isolated environment, including setting the environment variables they depend on, meaning we should set `ENV['ADMIN_HTTP_API_KEY']` in each test that requires it

Also if it's set to `supersecret` all the time when valid then we have no triangulation happenig

danielburnley: Additionally, I'm not entirely convinced this is the right way to do this, although it reduces duplication it also hides away from the tests what it's actually doing. As a person reading the `web_routes` tests it is now harder for me to see that you're setting the `API_KEY` header on the request, and I would have to navigate away from them to find this and see whats being set

danielburnley: Imo I would move this setup back into the tests that rely on it to avoid this hidden abstraction and make it easier for future people to find the setup

spec/admin_context.rb
```diff
@@ -0,0 +1,11 @@
+RSpec.shared_context 'with admin privileges' do
```
danielburnley: This should be done in each test for clarity - As a maintainer I don't want to be confused as to where to find how we set the auth headers.

The amount of duplication this reduces is very minimal and doesn't really provide much value in moving it into a globally shared context IMHO. As a maintainer, I'd have to open the tests, see some method I don't know where it comes from, look around for the context, and find this. 

I think there's value in actually setting these things explicitly in each test, it makes it much clearer what the setup is, and what the setup of the actual request it (i.e. I can see there's a header called API_KEY set)

This would involve moving this code back into `spec/web_routes/add_users_to_project_spec.rb`, and placing it inside your new test as well.

PR #58, opened 2018-09-10
Guard project find

goingdotin:
This PR fixes a typo in function that guards access to the endpoints.
It also adds guard to `project/find` endpoint
It also slightly refactors tests to make them (subjectively) more idiomatic.


faissalMT:
Obsoleted

PR #67, opened 2018-09-12
Expire API Tokens after 4 hours

danielburnley:
WHY - Prevents token reuse without having to log in again


spec/unit/local_authority/use_case/check_api_key_spec.rb
```diff
@@ -2,7 +2,8 @@
   let(:use_case) { described_class.new }
 
   def api_key_for_project(project, api_key=ENV['HMAC_SECRET'])
-    JWT.encode({project_id: project}, api_key, 'HS512')
+    expiry = (Time.now.to_i + 3600) * 3600
```
faissalMT: As someone reading this I don't know how much '3600' is

danielburnley: I found a better way of doing this in another test, i'll try do that :thinking: I hate working with timestamps

lib/local_authority/use_case/create_api_key.rb
```diff
@@ -1,6 +1,9 @@
+require 'date'
+
 class LocalAuthority::UseCase::CreateApiKey
   def execute(project_id:)
-    api_key = JWT.encode({project_id: project_id}, ENV['HMAC_SECRET'], 'HS512')
+    exp = DateTime.now.strftime("%s").to_i + (60 * 4)
```
mrosel: This is hardcoding an expiration - there are several benefits in doing this the other way around. I.e. have a `created_at` added to the JWT and use business logic to determine if it's expired. 
```
expires_in = 1.minute
expired = created_at > (Time.now - expires_in)
```

mrosel: persuaded by @danielburnley  to keep as is

spec/unit/local_authority/use_case/check_api_key_spec.rb
mrosel: Suggest that as `.to_i` also converts to unixtime, It's often easier to read by doing your datetime math, then converting.

mrosel: `ActiveSupport::Duration` could be used for readablity and conversion - https://api.rubyonrails.org/v4.2/classes/ActiveSupport/Duration.html

lib/local_authority/use_case/create_api_key.rb
robinlacey: First read I thought 60*4 was 4 minutes. Perhaps pull out to make a bit more readable?

spec/unit/local_authority/use_case/create_api_key_spec.rb
```diff
@@ -45,5 +67,24 @@
       
       expect(decoded_key[0]['project_id']).to eq(5)
     end
+
+    it 'Sets the expiry to 4 hours away' do
```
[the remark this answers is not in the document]
danielburnley: This is a duplicate from the two separate examples for triangulation purposes, the tests are duplicated in separate context blocks

PR #68, opened 2018-09-12
Add dependencies to return schema

robinlacey:


robinlacey:
Does not work with nested dependencies. Modified return schema.

robinlacey:
Merging. We all worked on this so there should be enough 👀

PR #75, opened 2018-09-18
Multiple dependencies

faissalMT:
What - Fixes the issues copying data to multiple dependencies in return populator


danielburnley:
As a future note - It'd be good to avoid having linting commits if you're going to use something like Rubocop locally, you can squash them into your previous commit to keep adhereing to the meaningful commits

faissalMT:
There's a very large refactoring that I have yet to get around to and that's breaking up all of the type stuff into its own use case. I think it's going to help significantly.

PR #78, opened 2018-09-19
LA-AC baseline template

robinlacey:


lib/homes_england/builder/template/template_builder.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class HomesEngland::Builder::Template::TemplateBuilder
+  def build_template(type:)
+    if type == 'hif'
+      HomesEngland::Builder::Template::Templates::HIFTemplate.new.create
```
danielburnley: Not sure these namespaces are the best for this - but probably worth discussing tomorrow

PR #83, opened 2018-09-20
Add /baseline/:type webroute

faissalMT:


lib/delivery_mechanism/web_routes.rb
```diff
@@ -19,6 +19,13 @@ class WebRoutes < Sinatra::Base
       200
     end
 
+    get '/baseline/:type' do
```
danielburnley: We'll want to `guard_access` this, just using the standard one will be fine for now, wouldn't worry about user roles just yet

faissalMT: The standard one will not work because it's been created under the assumption that users are tied to project ids, see:  
https://github.com/homes-england/monitor-api/blob/b35bcbd68c72b25b39f5ba80a2d14ff17586e718/lib/delivery_mechanism/web_routes.rb#L277

faissalMT: Unless you're suggesting passing it an id and doing nothing with it... That might work actually.

PR #84, opened 2018-09-21
Extract path typing functionality from PopulateReturnTemplate into GetReturnTemplatePathTypes

faissalMT:


lib/local_authority/use_case/get_return_template_path_types.rb
```diff
@@ -0,0 +1,104 @@
...
+
+  def get_path_types_from_array_last_dependency(path, schema)
+    schema.dig(:items, :dependencies).values.each do |value|
+      path_types = get_state_path(value, path)
+      return [:array] + path_types if path_found?(path_types)
+    end
+  end
+
+  def array_has_item_dependencies?(schema)
+    Common::HashUtils.hash_has_path(schema, %i[items dependencies])
```
danielburnley: You could also do this with a refinement (https://github.com/homes-england/monitor-api/blob/e3f251851de6093e0d5b369d35e1d9b0242769b6/lib/local_authority/refinement/deep_copy.rb), not necessary but worth looking at

lib/local_authority/use_case/get_return_template_path_types.rb
```diff
@@ -0,0 +1,104 @@
...
+    [:unsupported_type]
+  end
+
+  def get_schema_types_from_array(path, schema)
+    acquired_path = [:not_found]
+    if array_has_property?(path, schema)
+      acquired_path = get_path_types_from_array_properties(path, schema)
+    end
+    if array_has_item_dependencies?(schema) && path_not_found?(acquired_path)
+      acquired_path = get_path_types_from_array_last_dependency(path, schema)
```
danielburnley: > from array last dependency

Should that be `last dependency` still?

lib/local_authority/use_case/get_return_template_path_types.rb
faissalMT: Sounds good to me

lib/local_authority/use_case/get_return_template_path_types.rb
faissalMT: No

PR #85, opened 2018-09-25
Update baseline & return schema

danielburnley:
WHAT 
- Updates the baseline to allow for easier creation of the baseline editor in future
- Update the financials section to make it more clear to use


lib/homes_england/gateway/in_memory_template.rb
```diff
@@ -191,35 +259,63 @@ def hif_infrastructures
...
+                        title: 'Target date of planning granted'
+                      },
+                      summaryOfCriticalPath: {
+                        type: 'string',
+                        title: 'Summary of Critical Path'
+                      }
+                    }
+                  },
+                  {
+                    properties: {
```
robinlacey: Could just have default NA and it'll show nothing?

PR #86, opened 2018-09-25
Make /return/validate provide pretty paths

faissalMT:


danielburnley:
Would this work for something with a schema with two entries in the `oneOf` array in the dependencies?

faissalMT:
Yes, most of the code is copied from the latest version of GetSchemaCopyPaths.

PR #101, opened 2018-10-02
Rewrite the validation system to be less fragile

faissalMT:


spec/acceptance/local_authority/validate_return_spec.rb
```diff
@@ -18,8 +18,8 @@
     it 'should return invalid if fails validation' do
       valid_return = get_use_case(:validate_return).execute(type: 'hif', return_data: project_base_return_invalid)
       expect(valid_return[:valid]).to eq(false)
-      expect(valid_return[:invalid_paths]).to eq([[:infrastructures, 0, :planning, :outlinePlanning, :planningSubmitted, :percentComplete]])
-      expect(valid_return[:pretty_invalid_paths]).to eq([['HIF Project', 'Infrastructures', 'Infrastructure 1', 'Planning', 'Outline Planning', 'Planning Permission Submitted', 'Percent Complete']])
+      expect(valid_return[:invalid_paths]).to eq([[:infrastructures, 0, :planning, :outlinePlanning]])
+      expect(valid_return[:pretty_invalid_paths]).to eq([['HIF Project', 'Infrastructures', 'Infrastructure 1', 'Planning', 'Outline Planning']])
```
danielburnley: If there's two fields in there which are invalid, does that return multiple paths to the same place?

spec/unit/homes_england/builder/template/template_builder_spec.rb
```diff
@@ -4,7 +4,6 @@
   let(:template_builder) { described_class.new }
 
   it 'returns template object when hif requested' do
-    p template_builder
```
danielburnley: :eyes:

spec/acceptance/local_authority/validate_return_spec.rb
faissalMT: No, because the way oneOfs work is just trying to match states not fields, so we error once when we can't find a state that matches.

PR #106, opened 2018-10-03
Add titles to baseline

danielburnley:
Add titles to baseline template to allow the frontend baseline viewer to display things more clearly


lib/homes_england/builder/template/templates/hif_template.rb
```diff
@@ -93,134 +93,6 @@ def create()
     hif_template
   end
 
-  def hif_finances
```
faissalMT: Why is finances being removed?

danielburnley: We don't use that function anymore, it's the old finances section

PR #117, opened 2018-10-10
Create a status for projects

elcorbs:
You can now have a draft or submitted status on a project to signify when it's finished being edited.


spec/web_routes/submit_project_spec.rb
```diff
@@ -0,0 +1,44 @@
+# frozen-string-literal: true
+
+require 'rspec'
+require_relative 'delivery_mechanism_spec_helper'
+
```
faissalMT: `guard_access()` also provides a 401 if the API key is invalid, it would be a good idea to test this like so: https://github.com/homes-england/monitor-api/blob/master/spec/web_routes/create_return_spec.rb#L58

PR #119, opened 2018-10-11
Remove type from project update web route

elcorbs:


lib/delivery_mechanism/web_routes.rb
```diff
@@ -215,7 +215,6 @@ def invalid_validation_hash(request_hash:)
           update_successful = use_case.execute(
             id: request_hash[:id].to_i,
             project: {
-              type: request_hash[:project][:type],
               baseline: request_hash[:project][:baselineData]
```
faissalMT: Could you change this so it doesn't have to be nested? I'd like to just send `{id, data}` to this end point.

PR #120, opened 2018-10-15
Outputs forecast tab

danielburnley:
**WHAT**
- Updates baseline to have strings instead of integers
  - Migrates currently existing baselines
- Adds output forecast to return template

**WHY**
- Population doesn't work with integers, and there's currently no benefit to storing them as not strings
- Outputs forecast tab is required for future returns

### DIFF
--- a/spec/unit/database_administrator/migration_8_spec.rb
+++ b/spec/unit/database_administrator/migration_8_spec.rb
@@ -1,0 +1,180 @@
+# frozen_string_literal: true
+
+describe 'Migration 8' do
+  include_context 'with database'
+
+  def synchronize_to_non_migrated_version
+    migrator.migrate_to(database, 7)
+  end
+
+  def synchronize_to_migrated_version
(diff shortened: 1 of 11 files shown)


spec/unit/database_administrator/migration_8_spec.rb
```diff
@@ -0,0 +1,180 @@
+# frozen_string_literal: true
+
+describe 'Migration 8' do
+  include_context 'with database'
+
+  def syncronize_to_non_migrated_version
```
faissalMT: *synchronize

spec/unit/database_administrator/migration_8_spec.rb
```diff
@@ -0,0 +1,180 @@
...
+    migrator.migrate_to(database, 7)
+  end
+
+  def syncronize_to_migrated_version
```
faissalMT: Here too

PR #121, opened 2018-10-15
Notify tests

faissalMT:
What - Adds an acceptance spec for sending notifications to all users as well as refactoring some related tests.  
Why - This is needed to begin adding notifications on project creation (as of this notifications are only done for returns)

### DIFF
--- a/spec/simulator/notification.rb
+++ b/spec/simulator/notification.rb
@@ -1,0 +1,39 @@
+module Simulator
+  class Notification
+    extend Forwardable
+
+    def_delegators :@test_suite, :expect, :eq, :have_been_made, :hash_including
+
+    def initialize(test_suite, url)
+      @test_suite = test_suite
+      @url = url
+    end
+
+    def send_notification(to:)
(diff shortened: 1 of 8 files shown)


spec/simulator/notification.rb
```diff
@@ -0,0 +1,42 @@
...
+        request_body = JSON.parse(req.body)
+        expect(request_body["email_address"]).to eq(email_address)
+        expect(request_body["personalisation"]["access_url"]).to eq(access_url)
+      end).to have_been_made
+    end
+
+    def expect_notifier_to_be_accessed_with(url:)
+      expect(@stub_request.with do |req|
+        request_body = JSON.parse(req.body)
+        expect(req.uri.host.to_s).to eq(url)
```
danielburnley: The stub request itself has the URL in it (see line 15), so you don't need to worry about matching it inside here, you can just do:

`expect(@stub_request).to have_been_made`, which matches the URL it was initially set up with

danielburnley:
Might want to rebase this entire set of commits

PR #122, opened 2018-10-16
Create use case for validating new project data.

elcorbs:
Created Use Case, added a webroute.
Also moved use case for getting path titles to common as this was needed for both validating returns and projects.


lib/homes_england/domain/template.rb
```diff
@@ -1,3 +1,52 @@
+require 'json_schema'
 class HomesEngland::Domain::Template
   attr_accessor :layout, :schema
+
+  def invalid_paths(project_data)
```
danielburnley: Can some of these be private? (Not sure if they're used anywhere else)

PR #127, opened 2018-10-18
Business intelligence data export

faissalMT:


lib/homes_england/use_case/compile_project.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def execute(project_id:)
+    baseline = @find_project.execute(id: project_id)
+    baseline[:project_id] = project_id
+    baseline.delete(:status)
+
+    submitted_returns = @get_returns.execute(project_id: project_id)[:returns]
+    unless submitted_returns.empty?
+      submitted_returns.each do |i|
+        i[:data] = i.dig(:updates, -1)
```
danielburnley: This reads really unclearly to me, I assume you're getting the last update? 

It'd be easier to read if this do block was extracted out into a method such as `get_submitted_returns(returns)`, and additionally if that created its own hash and pulled the data out, rather than mutating and deleting data in the original hash (mutable data is not a fun thing to debug)

lib/homes_england/use_case/compile_project.rb
```diff
@@ -0,0 +1,28 @@
+class HomesEngland::UseCase::CompileProject
```
danielburnley: `CompileProject` doesn't reveal its intent clearly imo, when I see `Compile` I would think of a compiler - From the users perspective, what is this use case doing?

lib/homes_england/use_case/compile_project.rb
```diff
@@ -0,0 +1,28 @@
...
+    @find_project = find_project
+    @get_returns = get_returns
+  end
+
+  def execute(project_id:)
+    baseline = @find_project.execute(id: project_id)
+    baseline[:project_id] = project_id
+    baseline.delete(:status)
+
+    submitted_returns = @get_returns.execute(project_id: project_id)[:returns]
```
danielburnley: These aren't guaranteed to be submitted, right? This would just be all the returns not just the draft ones

faissalMT: I mistakenly thought it only returned submitted returns.

lib/homes_england/use_case/compile_project.rb
faissalMT: Combining all data associated with a project into a single hash.

lib/delivery_mechanism/web_routes.rb
```diff
@@ -132,6 +132,22 @@ class WebRoutes < Sinatra::Base
       end
     end
 
+    get '/project/:id/export' do
+      response.body = {}.to_json
+      guard_bi_access env, params, request do |request_hash|
+        compiled_project_hash = @dependency_factory.get_use_case(:export_project_data).execute(
```
danielburnley: `exported_project`?

lib/homes_england/use_case/compile_project.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def format_baseline_data(baseline_data, project_id)
+    new_baseline_data = baseline_data.dup
+    new_baseline_data[:project_id] = project_id
+    new_baseline_data.delete(:status)
+    new_baseline_data
+  end
+
+  def format_return_data(return_data)
+    new_return_data = return_data.dup
```
danielburnley: Would it be easier to simply create a new hash and set `:data` equal to the last update of the return data? It's clearer to the reader to see what you're explicitly grabbing, as opposed to what you're deleting

PR #128, opened 2018-10-18
Baseline Editor state

ccp92:
Adds a new state to projects of LA Draft.
Flow should now be:
- A new empty baseline is added: Status = Draft
- HE user adds information to the empty baseline and submits: Status = LA Draft
- LA user adds information to the partially filled baseline and submits: Status = Submitted

### DIFF
--- a/lib/homes_england/gateway/sequel_project.rb
+++ b/lib/homes_england/gateway/sequel_project.rb
@@ -34,12 +34,7 @@ class HomesEngland::Gateway::SequelProject
     { success: updated.positive? }
   end
 
-  def submit(id:)
-    row = @database[:projects].where(id: id).first
-    if row[:status] == 'Draft'
-      @database[:projects].where(id: id).update(status: 'LA Draft')
-    else
-      @database[:projects].where(id: id).update(status: 'Submitted')
-    end
+  def submit(id:, status:)
+    @database[:projects].where(id: id).update(status: status)
   end
 end
--- a/lib/homes_england/gateway/sequel_project.rb
+++ b/lib/homes_england/gateway/sequel_project.rb
@@ -31,10 +31,15 @@ class HomesEngland::Gateway::SequelProject
                 status: project.status
               )
 
-    { success: updated > 0 }
+    { success: updated.positive? }
   end
 
   def submit(id:)
-    @database[:projects].where(id: id).update(status: 'Submitted')
+    row = @database[:projects].where(id: id).first
+    if row[:status] == 'Draft'
+      @database[:projects].where(id: id).update(status: 'LA Draft')
+    else
+      @database[:projects].where(id: id).update(status: 'Submitted')
+    end
   end
 end
--- a/lib/homes_england/use_case/update_project.rb
+++ b/lib/homes_england/use_case/update_project.rb
@@ -7,7 +7,7 @@ class HomesEngland::UseCase::UpdateProject
     current_project = @project_gateway.find_by(id: project_id)
     updated_project = HomesEngland::Domain::Project.new
     updated_project.data = project_data
-    if current_project.status == 'Submitted' || current_project.status == 'LA Draft'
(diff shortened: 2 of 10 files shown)


lib/homes_england/use_case/update_project.rb
```diff
@@ -4,9 +4,14 @@ def initialize(project_gateway:)
   end
 
   def execute(project_id:, project_data:)
+    current_project = @project_gateway.find_by(id: project_id)
     updated_project = HomesEngland::Domain::Project.new
     updated_project.data = project_data
-    updated_project.status = 'Draft'
+    if current_project.status == 'Submitted' || current_project.status == 'LA Draft'
```
craigjbass:
```suggestion
    if current_project.submitted? || current_project.local_authority_draft?
```

spec/acceptance/homes_england/submit_project_spec.rb
```diff
@@ -5,32 +5,64 @@
 
 describe 'Submitting a completed draft project' do
   include_context 'dependency factory'
+  context 'Whilst status is LA Draft' do
+    it 'Changes the status to submitted' do
+      project_baseline = {
```
craigjbass: this baseline seems duplicated - could this be extracted to a let or method `empty_baseline` ?

spec/acceptance/homes_england/submit_project_spec.rb
```diff
@@ -5,32 +5,64 @@
...
-      financial: {
-        total_amount_estimated: ''
       }
-    }
-    response = get_use_case(:create_new_project).execute(
-      type: 'hif', baseline: project_baseline
-    )
-    get_use_case(:submit_project).execute(project_id: response[:id])
-    submitted_project = get_use_case(:find_project).execute(id: response[:id])
+      response = get_use_case(:create_new_project).execute(
```
craigjbass: could you extract some methods to hide some of this wiring in the it blocks?

```ruby
it 'changes status to LA Draft' do
  project_id = create_new_empty_project
  submit_project(project_id)
  expect_project_status_to_eq('LA Draft', project_id)
end
```

lib/homes_england/use_case/update_project.rb
```diff
@@ -4,12 +4,27 @@ def initialize(project_gateway:)
   end
 
   def execute(project_id:, project_data:)
+    current_project = @project_gateway.find_by(id: project_id)
     updated_project = HomesEngland::Domain::Project.new
     updated_project.data = project_data
-    updated_project.status = 'Draft'
+    if submitted?(current_project) || la_draft?(current_project)
```
danielburnley: Seems weird that update cares about the status - given it's only updating I wouldn't expect to need this logic

danielburnley: You could update the update method to no longer care about status and then you wouldn't need this

lib/homes_england/gateway/sequel_project.rb
```diff
@@ -31,10 +31,15 @@ def update(id:, project:)
...
               )
 
-    { success: updated > 0 }
+    { success: updated.positive? }
   end
 
   def submit(id:)
-    @database[:projects].where(id: id).update(status: 'Submitted')
+    row = @database[:projects].where(id: id).first
+    if row[:status] == 'Draft'
```
danielburnley: This looks like business logic that should probably exist in the use case which calls this - you'd need to update this method to take the "submitted status" or something, and update the `submit project` use case

lib/homes_england/use_case/update_project.rb
ccp92: The update use case needs to know about status to deal with the edge case that one user hits submit, followed by the second person hitting update, it should be in the draft status, not submitted.

Changing this would require another PR to give the second user a warning/error screen stating that the form has already been submitted/changed.

PR #136, opened 2018-10-24
Softcode type

faissalMT:


lib/local_authority/gateway/sequel_return.rb
```diff
@@ -11,11 +11,18 @@ def create(new_return)
   end
 
   def find_by(id:)
-    row = @database[:returns].where(id: id).first
+    row = @database[:returns]
+    .select_append(Sequel[:returns][:id].as(:return_id))
+    .select_append(Sequel[:returns][:status].as(:return_status))
+      .where(Sequel.qualify(:returns, :id) => id)
```
danielburnley: Ooh Sequel.qualify, nice find

lib/local_authority/use_case/create_return.rb
```diff
@@ -1,21 +1,24 @@
...
     { id: created_return_id }
   end
 
   private
 
-  def create_return_for_project(project_id)
+  def create_return_for_project(project_id, type)
     return_object = LocalAuthority::Domain::Return.new.tap do |r|
       r.project_id = project_id
+      r.type = type
```
danielburnley: Do we need to save this? I would have thought it would be pulled from the project so isn't required in create return

faissalMT: No we don't, this is a left over from my initial attempts.

faissalMT: No, this is a leftover from my initial attempts.

PR #139, opened 2018-10-25
Add tabs from spreadsheet to return Schema

ccp92:
Adds the following tabs when environment variables are enabled:

- Outputs - Actuals
- s151 Claim
- Submission
- RM Monthly Catchup

### DIFF
--- a/lib/local_authority/gateway/in_memory_return_template.rb
+++ b/lib/local_authority/gateway/in_memory_return_template.rb
@@ -1715,7 +1715,8 @@ class LocalAuthority::Gateway::InMemoryReturnTemplate
 
     add_s151_tab
     add_outputs_actuals_tab
-    
+    add_rm_monthly_catchup_tab
+
     @return_template
   end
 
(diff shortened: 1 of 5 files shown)


lib/local_authority/gateway/in_memory_return_template.rb
```diff
@@ -1712,11 +1723,577 @@ def find_by(type:)
...
+                },
+                descriptionOfSpend: {
+                  title: 'Description of Spend',
+                  type: 'string',
+                  extendedText: true
+                },
+                evidenceOfSpendNextQuarter: {
+                  title: 'Evidence of Spend for the Past Quarter.',
+                  type: 'string',
+                  hiddem: true
```
danielburnley: > hiddem

lib/local_authority/gateway/in_memory_return_template.rb
```diff
@@ -1712,11 +1723,577 @@ def find_by(type:)
...
+              }
+            }
+          }
+        }
+      }
+    }
+  end
+
+  def add_rm_monthly_catchup_tab
+    return if ENV['RM_MONTHYL_CATCH_TAB'].nil?
```
danielburnley:
```suggestion
    return if ENV['RM_MONTHLY_CATCHUP_TAB'].nil?
```

PR #142, opened 2018-10-25
Add adaptors between LA/HE use cases and gateways

danielburnley:
WHAT
- Stops LA/HE use cases calling the gateways directly, instead calling a `ui` version of the use case
  - For this PR this simply passes the request directly to the gateway, in prep for a more specific adaptor in the future

WHY
- The database structure was too coupled to the front end schema, this visual changes on the front end (e.g. moving sections) difficult as they required DB migrations, this meant we couldn't quickly iterate on changes
- This provides the first step to allowing our frontend structure to be decoupled from our database structure


lib/local_authority/use_case/monitor_get_return_updates.rb
```diff
@@ -0,0 +1,16 @@
...
+
+class LocalAuthority::UseCase::MonitorGetReturnUpdates
+  def initialize(return_update_gateway:)
+    @return_update_gateway = return_update_gateway
+  end
+
+  def execute(return_id:)
+    updates = @return_update_gateway.updates_for(return_id: return_id)
+    {
+      updates:  updates.map do |u|
```
danielburnley: Would look nicer if moved out to a method

lib/local_authority/use_cases.rb
```diff
@@ -48,10 +48,16 @@ def self.register(builder)
       )
     end
 
+    builder.define_use_case :monitor_return_updates do
```
danielburnley: `monitor_return_updates` is a weird name :thinking:

lib/homes_england/use_cases.rb
```diff
@@ -8,9 +8,15 @@ def self.register(builder)
       )
     end
 
+    builder.define_use_case :monitor_get_project do
```
danielburnley: `monitor_get_project` is a weird name

craigjbass: Yes

danielburnley:
Closed - superseded by #152

PR #149, opened 2018-10-30
Add currency: true to monetary fields.

elcorbs:


lib/local_authority/gateway/in_memory_return_template.rb
```diff
@@ -2106,16 +2155,19 @@ def add_s151_tab
...
                 },
                 actual: {
                   title: 'Actual Spend Last Quarter Month',
-                  type: 'string'
+                  type: 'string',
+                  currency: true
                 },
                 varianceAgainstForcastAmount: {
                   title: 'Variance Against Forecast (£)',
                   type: 'string',
```
craigjbass: is it not possible to have `type: 'currency'` ?

craigjbass: (don't even know) cc @cybojenix

elcorbs: I've written a custom component to display a little currency symbol next to the input box and this is the marker for it. The actual number will still be a string, at least for now.

PR #154, opened 2018-10-31
Add layer between UI and core for returns

danielburnley:
WHAT

Adds usecases between the UI and the core application for returns

WHY

First steps towards allowing us to update the frontend schema without having to change the way it is internally stored

### DIFF
--- a/spec/unit/ui/use_case/create_project_spec.rb
+++ b/spec/unit/ui/use_case/create_project_spec.rb
@@ -39,7 +39,7 @@ describe UI::UseCase::CreateProject do
     end
   end
 
-  context 'Example one' do
+  context 'Example two' do
     let(:create_project_spy) { spy(execute: { id: 5 }) }
     let(:use_case) { described_class.new(create_project: create_project_spy) }
     let(:response) do
--- a/spec/unit/ui/use_case/create_project_spec.rb
+++ b/spec/unit/ui/use_case/create_project_spec.rb
@@ -1,0 +1,83 @@
+# frozen_string_literal: true
+
+describe UI::UseCase::CreateProject do
+  context 'Example one' do
+    let(:create_project_spy) { spy(execute: { id: 1 }) }
+    let(:use_case) { described_class.new(create_project: create_project_spy) }
+    let(:response) do
+      use_case.execute(type: 'hif', name: 'Cats', baseline: { Cats: 'purr' }) 
+    end
+
+    before do
+      response
+    end
+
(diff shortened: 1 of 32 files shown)


danielburnley:
Depends on #152

spec/unit/ui/use_case/create_project_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+        have_received(:execute).with(hash_including(baseline: { Cats: 'purr' }))
+      )
+    end
+
+    it 'Returns the id from create project' do
+      expect(response).to eq(id: 1)
+    end
+  end
+
+  context 'Example one' do
```
ccp92:
```suggestion
  context 'Example two' do
```
You've put example one twice.

spec/unit/ui/use_case/update_project_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+        have_received(:execute).with(
+          hash_including(
+            project_data: { dog: 'woof' }
+          )
+        )
+      )
+    end
+
+    it 'Returns successful if successful' do
+      expect(response).to eq(successful: false)
```
ccp92: This test is a little confusing?
It should return successful but we're expecting `successful: false`?
(Sorry, I'm probably missing something obvious)

spec/unit/ui/use_case/update_return_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+        )
+      )
+    end
+
+    it 'returns an empty hash' do
+      expect(response).to eq({})
+    end
+  end
+
+  context 'Example one' do
```
ccp92:
```suggestion
  context 'Example two' do
```

spec/unit/ui/use_case/update_project_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+      expect(update_project_spy).to(
+        have_received(:execute).with(
+          hash_including(
+            project_data: { dog: 'woof' }
+          )
+        )
+      )
+    end
+
+    it 'Returns successful if successful' do
```
danielburnley:
```suggestion
    it 'Returns unsuccessful if unsuccessful' do
```

spec/unit/ui/use_case/update_project_spec.rb
danielburnley: Badly worded test - have updated

PR #155, opened 2018-11-01
AC baseline improvements

ccp92:
Add some missing fields/ move some fields based on the source spreadsheet


lib/homes_england/builder/template/templates/ac_template.rb
```diff
@@ -20,182 +20,237 @@ def create
...
+                  conditionName: {
                     type: 'string',
-                    title: 'Category Title'
+                    title: 'Condition Name'
                   },
-                  percent: {
+                  conditionExplanation: {
                     type: 'string',
-                    title: 'Percent Amount'
+                    title: 'Condition explanation'
```
faissalMT: This should be titleized

PR #156, opened 2018-11-01
Add conversion methods for project related interactions

danielburnley:
WHAT
- Adds converts for hif projects that sit between the UI/Core layer

WHY
- To allow for easier changing of the schema for the UI without requiring DB changes

**OUTSTANDING WORK**
- Support for AC
  - As this is still in heavy development I have just ignored converting AC for the moment
- Schemas for the UI layer need to be returned over the internal schemas
  - as there is no difference yet this can be done in a subsequent PR

### DIFF
--- a/lib/ui/use_case/get_project.rb
+++ b/lib/ui/use_case/get_project.rb
@@ -1,12 +1,15 @@
 # frozen_string_literal: true
 
 class UI::UseCase::GetProject
-  def initialize(find_project:)
+  def initialize(find_project:, convert_core_hif_project:)
     @find_project = find_project
+    @convert_core_hif_project = convert_core_hif_project
   end
 
   def execute(id:)
     found_project = @find_project.execute(id: id)
+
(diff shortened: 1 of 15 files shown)


lib/ui/use_case/get_project.rb
```diff
@@ -1,18 +1,27 @@
...
+  def initialize(find_project:, convert_core_hif_project:)
     @find_project = find_project
+    @convert_core_hif_project = convert_core_hif_project
   end
 
   def execute(id:)
     found_project = @find_project.execute(id: id)
 
+    found_project[:data] = convert_data(found_project) unless found_project[:type] != 'hif'
+
```
elcorbs: Is there a double negative here for a reason? Why not ``` if found_project[:type] == 'hif' 
 ```?

elcorbs:
```suggestion
found_project[:data] = convert_data(found_project) if found_project[:type] == 'hif'
```

PR #157, opened 2018-11-01
Add UI specific schema

danielburnley:


lib/homes_england/use_case/get_schema_for_project.rb
```diff
@@ -5,6 +5,7 @@ def initialize(template_gateway:)
 
   def execute(type:)
     template = @template_gateway.get_template(type: type)
+    p template
```
elcorbs: Is this accidentally left in?

elcorbs:
```suggestion
```

PR #180, opened 2018-11-08
Change financials tab to match feedback

ccp92:
screenshot-localhost-3000-2018 11 08-16-39-48

### DIFF
--- a/lib/ui/gateway/schemas/ac_project.json
+++ b/lib/ui/gateway/schemas/ac_project.json
@@ -192,13 +192,13 @@
         "expenditure": {
           "type": "array",
           "addable": true,
-          "title": "Expenditure",
+          "title": "Funding Drawdown",
           "items": {
             "type": "object",
             "properties": {
               "fundingDrawdown": {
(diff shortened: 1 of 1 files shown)


lib/ui/gateway/schemas/ac_project.json
```diff
@@ -212,64 +212,37 @@
...
           "addable": true,
-          "title": "Expenditure",
+          "title": "Funding Drawdown",
           "items": {
             "type": "object",
             "properties": {
               "fundingDrawdown": {
                 "type": "object",
-                "title": "Funding Drawdown",
+                "title": "",
```
elcorbs: Should probably take them out completely instead of leaving empty strings?

ccp92: I saw this on one of Faissal's PRs and thought the same thing.
Unfortunately, if an object doesn't have a title, it uses the name of the object instead.
(e.g. in this example, the UI would show "fundingDrawdown" as the header rather than leave it blank.)

PR #181, opened 2018-11-09
Add percentage widget to percentage fields

ccp92:


lib/local_authority/gateway/in_memory_return_template.rb
```diff
@@ -1015,7 +1023,8 @@ def find_by(type:)
                         },
                         percentComplete: {
                           type: 'integer',
```
faissalMT: I don't believe the percentage widget expects an integer and we use string for pretty much everything so I think we should change this before merging.

PR #184, opened 2018-11-12
Add use cases to convert between ui and core data for returns

elcorbs:


ccp92:
Not sure if this will be up to date anymore as I think there were other PRs since then that modify the returns data structure?

PR #201, opened 2018-11-14
Only show variance fields if there is variance

faissalMT:
Uses radio buttons to allow the user to choose whether or not there is variance as opposed to conditional showing depending on entered value because this is not possible in JSON Schema v4.


spec/unit/ui/use_case/convert_ui_hif_return_spec.rb
```diff
@@ -1,4 +1,4 @@
-describe UI::UseCase::ConvertUIHIFReturn do
+fdescribe UI::UseCase::ConvertUIHIFReturn do
```
elcorbs:
```suggestion
describe UI::UseCase::ConvertUIHIFReturn do
```

PR #203, opened 2018-11-15
Fix error on converting project.

elcorbs:
Moved the compact inside the map and put some catches in to catch arrays of nils.

### DIFF
--- a/lib/ui/use_case/convert_ui_hif_project.rb
+++ b/lib/ui/use_case/convert_ui_hif_project.rb
@@ -151,23 +151,23 @@ class UI::UseCase::ConvertUIHIFProject
   end
 
   def convert_funding_profiles
-    @converted_project[:fundingProfiles] = []
-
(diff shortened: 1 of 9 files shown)


lib/ui/use_case/convert_ui_hif_project.rb
```diff
@@ -151,23 +151,23 @@ def convert_infrastructures
...
-        }
+      {
+        period: profile[:period],
+        instalment1: profile[:instalment1],
+        instalment2: profile[:instalment2],
+        instalment3: profile[:instalment3],
+        instalment4: profile[:instalment4],
+        total: profile[:total]
+      }
+      profile.compact
```
faissalMT: This would return the result of profile.compact, making the hash prior redundant.

elcorbs: Sorry that was silly, changed this now

PR #205, opened 2018-11-16
Add ac returns schema

elcorbs:
This will error if checking on the master branch currently, due to the Project Summary page. Can check it if you check out branch change-project-summary-to-allow-objects.


lib/local_authority/gateway/in_memory_return_template.rb
```diff
@@ -3,7 +3,2223 @@
...
   def find_by(type:)
-    return nil unless type == 'hif'
+    return create_hif_schema if type == 'hif'
+    return create_ac_schema if type == 'ac'
+    nil
+  end
+
+  private
+
+  def create_ac_schema
```
faissalMT: Would be good to split these schemas out into two files

PR #210, opened 2018-11-19
Fix a couple small errors

elcorbs:
- Fixed a type in convertors.
- Removed unnecessary dependencies causing a validation error.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -33,7 +33,6 @@ services:
       S151_TAB: 'yes'
       RM_MONTHLY_CATCHUP_TAB: 'yes'
       OUTPUTS_ACTUALS_TAB: 'yes'
-      NO_LOGS: 'true'
 
   notify_simulator:
     build: .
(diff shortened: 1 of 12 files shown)


docker-compose.yml
```diff
@@ -33,6 +33,7 @@ services:
       S151_TAB: 'yes'
       RM_MONTHLY_CATCHUP_TAB: 'yes'
       OUTPUTS_ACTUALS_TAB: 'yes'
+      NO_LOGS: 'true'
```
ccp92: Should probably take this back out again before merging into master

```suggestion
```

lib/common/domain/template.rb
```diff
@@ -28,6 +28,7 @@ def get_invalid_paths(schema, project_data)
     nil
   rescue JsonSchema::AggregateError => aggregated_errors
     get_supported_errors(aggregated_errors).map do |error|
+      pp error
```
ccp92: Should take out the pp

elcorbs:
```suggestion
```

PR #222, opened 2018-11-26
Use ui return schema when getting the base return 

elcorbs:
In the UI use case for getting the base return, make sure it returns the ui return schema to the get base return end point.

### DIFF
--- a/lib/ui/use_case/get_base_return.rb
+++ b/lib/ui/use_case/get_base_return.rb
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 
 class UI::UseCase::GetBaseReturn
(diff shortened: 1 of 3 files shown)


lib/ui/use_case/get_base_return.rb
```diff
@@ -13,6 +14,14 @@ def execute(project_id:)
     base_return = @get_base_return.execute(project_id: project_id)[:base_return]
 
     base_return[:data] = @convert_core_hif_return.execute(return_data: base_return[:data]) if type == 'hif'
-    { base_return: base_return }
+    base_return[:schema] = @return_schema.find_by(type: type).schema
+
+    {
+      base_return: {
```
faissalMT: Why not just continue returning `{ base_return: base_return }`?

PR #224, opened 2018-11-28
WIP: Add instructions for adding new users

ccp92:
WHAT:
Add instructions for adding new users to the system.

WHY:
The old method no longer works and this is not documented anywhere at the moment.


README.md
```diff
@@ -26,10 +26,26 @@ Once you have cloned the repository you can run the application with the followi
 
 The application runs on port `4567`
 
+## Adding Users
```
danielburnley: Isn't there also an API end point for adding users to projects? Or has this not been updated with regards to roles?

ccp92: I think from what @faissalMT mentioned yesterday the API endpoint won't work with the new roles and it will be a quite fiddly piece of work to adjust how it works.
Hopefully, he can confirm?

ccp92: #228 will allow the http method to include user role going forwards so this will be changed then.

ccp92:
Once #228 is merged, these instructions will need changing so hold off on merging.

ccp92:
Not sure if I should update this to include the http gem instructions after all?
Do we want to input instructions to the README that requires a third party gem that's not included with the repo to use?

PR #226, opened 2018-11-28
Combine Planning Status, Section 106 and Statutory Consents

ccp92:
WHAT:
Combines these three tabs under the infrastructure headings into one tab.

WHY:
Feedback from users was that there was no need for such small fields to be on different pages.

NEW:
screenshot-localhost-3000-2018 11 29-11-50-18

PREVIOUS:
screenshot-localhost-3000-2018 11 29-11-52-09

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -33,7 +33,6 @@ services:
       S151_TAB: 'yes'
       RM_MONTHLY_CATCHUP_TAB: 'yes'
       OUTPUTS_ACTUALS_TAB: 'yes'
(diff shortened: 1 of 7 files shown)


docker-compose.yml
```diff
@@ -33,6 +33,7 @@ services:
       S151_TAB: 'yes'
       RM_MONTHLY_CATCHUP_TAB: 'yes'
       OUTPUTS_ACTUALS_TAB: 'yes'
+      NO_LOG: 'true'
```
```suggestion
```

PR #228, opened 2018-11-29
Change the add user end point to except user roles.

elcorbs:
So that we can add users with roles to projects by calling the add users endpoint with:   { users: [ { email: "myemail@ahome.com, role: "A Role" }] }

### DIFF
--- a/lib/homes_england/use_case/add_user_to_project.rb
+++ b/lib/homes_england/use_case/add_user_to_project.rb
@@ -3,11 +3,12 @@ class HomesEngland::UseCase::AddUserToProject
     @user_gateway = user_gateway
   end
 
-  def execute(email:, project_id:)
+  def execute(email:, role: nil, project_id:)
     user = @user_gateway.find_by(email: email)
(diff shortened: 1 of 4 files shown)


lib/homes_england/use_case/add_user_to_project.rb
```diff
@@ -3,11 +3,12 @@ def initialize(user_gateway:)
     @user_gateway = user_gateway
   end
 
-  def execute(email:, project_id:)
+  def execute(email:, role: nil, project_id:)
     user = @user_gateway.find_by(email: email)
     if user.nil?
       user = LocalAuthority::Domain::User.new.tap do |u|
```
faissalMT: Doesn't the AddUser usecase do this?

elcorbs: Yeah but this is what we use to add people to a project and links to /project/id/add_user, so now we can do it in one go instead  of adding users on cli then adding them to the project.

PR #233, opened 2018-11-30
Add link to be sent to gov uk notification gateway

elcorbs:


spec/unit/local_authority/gateway/gov_email_notification_gateway_spec.rb
```diff
@@ -108,8 +108,8 @@
           simulator.expect_notifier_to_have_been_accessed
         end
 
-        it 'will run send_email with address and url within personalisation hash' do
-          simulator.expect_notification(email_address: 'cat@cathouse.com', personalisation: {access_url: 'http://cats.com', by: 'Cat', project_name: 'Flap'})
+        it 'will run send_email with address, url and print urlwithin personalisation hash' do
```
ccp92:
```suggestion
        it 'will run send_email with address, url and print url within personalisation hash' do
```

PR #237, opened 2018-12-03
Add fields around recovery of funds HIF baseline

elcorbs:


lib/ui/gateway/schemas/hif_project.json
```diff
@@ -577,46 +577,89 @@
...
+                },
+                "dependencies": {
+                  "aimToRecover": {
+                    "oneOf": [
+                      {
+                        "properties": {
+                          "aimToRecover": {"enum": ["Yes"]},
+                          "expectedAmount": {
+                            "type": "string",
+                            "title": "Expected Amount"
```
ccp92: Might want to add `"currency": true` here?

PR #239, opened 2018-12-03
Add timestamp to project data

elcorbs:
- Adds a timestamp to the projects table
 - It will send this timestamp out in find project
 - When the frontend sends an update request, it includes the timestamp in the data; it checks the timestamp in the database against this timestamp.
 - If the timestamps are different it won't save the data and will return an error. 

This is to stop a later update request, using out of date data, from overwriting a more recent copy of the data.

This is to stop two people who have the project open saving one after the other and the second one saving overwriting the first ones save.

### DIFF
--- a/spec/unit/homes_england/use_case/update_project_spec.rb
+++ b/spec/unit/homes_england/use_case/update_project_spec.rb
@@ -3,94 +3,192 @@
 require 'rspec'
 
 describe HomesEngland::UseCase::UpdateProject do
-  let(:use_case) { described_class.new(project_gateway: project_gateway_spy) }
-  let(:response) { use_case.execute(project_id: project_id, project_data: updated_project_data) }
-  let(:la_project) { HomesEngland::Domain::Project.new }
+  context 'first update' do 
+    let(:use_case) { described_class.new(project_gateway: project_gateway_spy) }
+    let(:response) do
+      use_case.execute(
+        project_id: project_id,
+        project_data: updated_project_data,
+        timestamp: timestamp
+      )
+    end
+    let(:la_project) { HomesEngland::Domain::Project.new }
+  
+    before do
+      response
+    end
+  
+  
+    context 'example one' do
+      let(:project_id) { 42 }
 
-  before do
-    response
-  end
-
-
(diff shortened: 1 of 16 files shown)


spec/unit/homes_england/use_case/update_project_spec.rb
```diff
@@ -3,94 +3,192 @@
...
+  
+        let(:project_gateway_spy) do
+          spy(find_by: la_project, update: { success: true })
+        end
 
-      it 'Should pass the project to the gateway' do
-        expect(project_gateway_spy).to have_received(:update) do |request|
-          project = request[:project]
-          expect(project.data).to eq(ducks: 'quack')
+        it 'pass the data to the gateway' do
```
danielburnley:
```suggestion
        it 'does not pass the data to the gateway' do 
```

spec/unit/homes_england/use_case/update_project_spec.rb
```diff
@@ -3,94 +3,192 @@
...
-      let(:project_gateway_spy) do
-        spy(find_by: la_project, update: { success: true })
-      end
 
-      it 'Should pass the ID to the gateway' do
-        expect(project_gateway_spy).to have_received(:update).with(
-          hash_including(id: 123)
-        )
-      end
+        it 'pass the data to the gateway' do
```
danielburnley:
```suggestion
        it 'does not pass the data to the gateway' do 
```

spec/unit/homes_england/use_case/update_project_spec.rb
```diff
@@ -3,94 +3,192 @@
...
+          p.status = 'Draft'
+          p.data = { a: 'b' }
+          p.timestamp = time_now
         end
+  
+      project_gateway_spy = spy(find_by: current_project, update: { success: true })
+      
+      use_case = described_class.new(project_gateway: project_gateway_spy)
+
+      sleep(1)
```
danielburnley: eesh

elcorbs: thought you might hate that..

PR #240, opened 2018-12-03
Add unit calculations

faissalMT:


elcorbs:
As the calculations are marked in the object above, If there are two calculations in the same object how would you mark this in the schema. Like how would mark a new line in the eval?

faissalMT:
You can either use a `;` or `\n` but I'd prefer the former.

lib/ui/gateway/schemas/ac_project.json
```diff
@@ -78,6 +78,7 @@
               },
               "units": {
                 "type": "object",
+                "calculation": "formData.numberOfUnitsTotal = formData.numberOfUnits ? parseMoney(formData.numberOfUnits.numberOfUnitsMarket) + parseMoney(formData.numberOfUnits.numberOfUnitsSharedOwnership) + parseMoney(formData.numberOfUnits.numberOfUnitsAffordable) + parseMoney(formData.numberOfUnits.numberOfUnitsPRS) + parseMoney(formData.numberOfUnits.numberOfUnitsOther): 0;",
```
craigjbass: This reverse couples the api to the react Frontend.

PR #242, opened 2018-12-04
Add CircleCI config ready for GovPaaS migration

davidwinter:
The config for CircleCI is located in `.circleci/config.govpaas.yml` and can we swapped out when ready to migrate across. This will not interfere with the current CircleCI configuration.

It's tricky to test currently as we don't have a paid account for GovPaaS, so some tweaking might be required when ready to migrate, such as memory of instances. Secrets will also need adding to CircleCI when ready, and finally the DB migrated from Heroku.

### DIFF
--- a/scripts/flip-dark-to-live.sh
+++ b/scripts/flip-dark-to-live.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+set -e
+
+if [ -z "$1" ]; then
+  echo 'Provide an environment to deploy'
+  exit 1
+fi
+
+ENVIRONMENT_NAME="$1"
+SPACE_NAME="${ENVIRONMENT_NAME}"
+APP_NAME="monitor-api-${ENVIRONMENT_NAME}"
+APP_HOSTNAME="monitor-api-${ENVIRONMENT_NAME}"
+DARK_APP_NAME="${APP_NAME}-dark"
+DARK_APP_HOSTNAME="${APP_NAME}-dark"
+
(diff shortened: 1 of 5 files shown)


scripts/flip-dark-to-live.sh
```diff
@@ -0,0 +1,31 @@
...
+
+./cf target -o "${CF_ORG}" -s "${SPACE_NAME}"
+
+./cf map-route "${DARK_APP_NAME}" "${CF_DOMAIN}" -n "${APP_HOSTNAME}"
+./cf unmap-route "${APP_NAME}" "${CF_DOMAIN}" -n "${APP_HOSTNAME}"
+
+./cf map-route "${APP_NAME}" "${CF_DOMAIN}" -n "${DARK_APP_HOSTNAME}"
+./cf unmap-route "${APP_NAME}-dark" "${CF_DOMAIN}" -n "${DARK_APP_HOSTNAME}"
+
+./cf rename "${APP_NAME}" "${APP_NAME}-temp"
```
cybojenix: without checking the docs, is this how cloudfoundry blue green works?
this looks like it will cause downtime.

Or are they doing some hidden magic here to keep trickling to the `temp` instance?

davidwinter: The traffic is already being routed to the dark instance above which is then in effect live. This is blue/green for Cloud Foundry.

https://circleci.com/docs/2.0/deployment-integrations/#live-deployment
https://docs.cloudfoundry.org/devguide/deploy-apps/blue-green.html

cybojenix: ah I missed the map routes

PR #247, opened 2018-12-04
WIP Enables mini profiler

booyaa:
Don't merge, creating PR to open discussion about the approach.

**What**

Enable mini-profiler for API. Uses a sidecar docker compose service to view trace files.

**Why**

Mini-profiler expects your ruby app to have at least one web page so it can inject the profiler widget into the page. To work around this with an API service, we need to store the trace files in a common directory (`./tmp/profiler`) and point our sidecar service to view the trace files.

A better solution would involve a mini profiler cli. This was mooted, but never acted upon. https://github.com/MiniProfiler/rack-mini-profiler/issues/365


booyaa:
Suboptimal solution, alternative COTS needs to be investigated i.e. New Relic

PR #250, opened 2018-12-05
Add specific lint targets

booyaa:
**What**

Adds specific lint targets to assess style or complexity related linting violations.

**Why**

Will allow developers to work on a specific set of violations.


cybojenix:
While this is useful for getting over the initial hurdle of mass changes, it isn't something to be used afterwards.
May be worth talking to the rest of the team to see how they want to tackle the current 3k warnings, and propose this as a solution.

PR #255, opened 2018-12-07
WIP: Add variance calculations where required

ccp92:


ccp92:
Needs to have a change made to the horizontal component to allow it to work with calculations.

PR #261, opened 2018-12-11
Extend source key so that baseline data can be chosen if there is no return data present.

elcorbs:
- changed the funding profile to quarterly components
- added a hidden profile to chose either previous profile or new profile that was requested
- extend source key so that you can decided to default to baseline if no previous return
- Use this on funding request to pull from baseline if no previous return, if not the hidden profile of previous return.

### DIFF
--- a/lib/local_authority/gateway/hif_returns_schema.rb
+++ b/lib/local_authority/gateway/hif_returns_schema.rb
@@ -1283,7 +1283,7 @@ class LocalAuthority::Gateway::HIFReturnsSchemaTemplate
                     period: {
                       title: 'Period',
                       type: 'string',
-                      sourceKey: %i[baseline_data fundingProfiles period],
+                      sourceKey: [:return_or_baseline, [:baseline_data, :fundingProfiles, :period],[:return_data, :fundingProfiles, :currentFunding, :forecast, :period]],
                       readonly: true
                     },
                     forecast: {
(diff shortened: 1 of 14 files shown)


lib/local_authority/gateway/hif_returns_schema.rb
```diff
@@ -1331,6 +1331,58 @@ def execute
                   }
                 }
               },
+              currentFunding: {
+                type: "object",
+                title: " ",
```
ccp92: Having these as whitespace characters causes extra sections to appear in the form, even when the dependency hasn't been selected yet.
```suggestion
                title: "",
```

lib/local_authority/gateway/hif_returns_schema.rb
```diff
@@ -1331,6 +1331,58 @@ def execute
...
+                properties: {
+                  forecast: {
+                  title: " ",
```
ccp92: Same as above
```suggestion
                  title: "",
```

lib/local_authority/use_case/find_path_data.rb
```diff
@@ -1,5 +1,14 @@
 class LocalAuthority::UseCase::FindPathData
   def execute(baseline_data:, path:)
+
+    if path[0] == :return_or_baseline
```
ccp92: Maybe extract this into a private method to be called be execute to make it clearer in the future what it's doing?

PR #277, opened 2018-12-17
Move fields to their correct dependencies

ccp92:
Some of the s106 options weren't appearing in dependencies when they should have been.


ccp92:
Would be good to have some feedback on this as moving the fields into dependencies seems to have stopped the readonly tag from working...

PR #282, opened 2018-12-18
Change title and add description

elcorbs:


ccp92:
I've approved this but just a heads up that I think the titles in the variance component are hardcoded at the moment so this won't make a visible difference on the front end.

PR #301, opened 2019-01-02
Procurement variance and dependencies

faissalMT:
Cards 71, 70, 72 & 65


elcorbs:
When you select completed, shouldn't you be able to enter the 'Completed Date' and 'Completion Name of Contractor' instead of it being readonly only and calculated?

PR #304, opened 2019-01-03
Make statutory consents independent of section 106 status

faissalMT:
Card 66 & 67
image
image
image
image

### DIFF
--- a/lib/local_authority/gateway/hif_returns_schema.rb
+++ b/lib/local_authority/gateway/hif_returns_schema.rb
@@ -494,7 +494,8 @@ class LocalAuthority::Gateway::HIFReturnsSchemaTemplate
                                           currentReturn: {
                                             title: 'Current return',
                                             type: 'string',
-                                            format: 'date'
+                                            format: 'date',
+                                            sourceKey: %i[return_data infrastructures planning section106 statutoryConsents statutoryConsents currentReturn]
                                           },
                                           previousReturn: {
                                             title: 'Previous Return',
@@ -1857,7 +1858,7 @@ class LocalAuthority::Gateway::HIFReturnsSchemaTemplate
                 quarter4:{
                   type: "string",
                   title: "Q4 Current Year"
-                }  
+                }
               }
             },
             actualAmounts: {
@@ -1880,7 +1881,7 @@ class LocalAuthority::Gateway::HIFReturnsSchemaTemplate
                 quarter4:{
                   type: "string",
                   title: "Q4 Current Year"
-                }  
+                }
               }
             },
             progress: {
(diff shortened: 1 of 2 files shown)


lib/ui/gateway/schemas/hif_return.json
```diff
@@ -466,90 +466,101 @@
...
-                                      "title": "Variance against last return (Calculated)",
+                                      "percentage": true
+                                    }
+                                  }
+                                },
+                                {
+                                  "properties": {
+                                    "statusAgainstLastReturn": { "enum": ["Completed"] },
+                                    "completionDate": {
+                                      "title": "Completion date (Calculated)",
```
elcorbs: This isn't calculated

lib/local_authority/gateway/hif_returns_schema.rb
```diff
@@ -494,7 +494,8 @@ def execute
                                           currentReturn: {
                                             title: 'Current return',
                                             type: 'string',
-                                            format: 'date'
+                                            format: 'date',
+                                            sourceKey: %i[return_data infrastructures planning section106 statutoryConsents statutoryConsents currentReturn]
```
elcorbs: Should this sourceKey not be on previous return?

lib/ui/gateway/schemas/hif_return.json
```diff
@@ -558,28 +564,48 @@
...
-                                      "title": "Reason for Variance",
-                                      "type": "string"
-                                    },
-                                    "percentComplete": {
-                                      "title": "Percentage Complete",
+                                    "varianceAgainstBaseline": {
+                                      "title": "Variance Against Baseline (weeks) (Calculated)",
                                       "type": "string",
-                                      "percentage": true
+                                      "readonly": true
```
elcorbs: This readonly thing isn't working for me. Although probably not an issue for this PR.

faissalMT: Yeah alot of our custom fields aren't working for dependencies, I've put it under card 121

lib/local_authority/gateway/hif_returns_schema.rb
faissalMT: No because return_data refers to the previous return, maybe it would be a good idea to consider renaming it at some point.

elcorbs: I mean should the source key not be in the previous return field instead of the current return field. So in the current return you enter a new value.

faissalMT: It's also in the previous return field, it's just being used as a default here so if it's shown and the user overrides it it'll be what's used in the next return. Currently if you create a return, fill in currentReturn, create another return, don't fill it in and then create another return the value in previousReturn won't show the date.

PR #305, opened 2019-01-04
Funding packages adjustments

elcorbs:
- Added description of infrastructure
- Added variance calculations to total cost and hif spend section
- Added some missing fields

### DIFF
--- a/lib/ui/gateway/schemas/hif_return.json
+++ b/lib/ui/gateway/schemas/hif_return.json
@@ -1838,7 +1838,7 @@
               "totalCost": {
                 "title": "",
                 "type": "object",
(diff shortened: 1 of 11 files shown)


lib/ui/gateway/schemas/hif_return.json
```diff
@@ -1822,76 +1822,150 @@
           "fundingStack": {
             "type": "object",
             "title": "Funding stack",
-            "calculation": "set(formData['totalCost'], 'percentComplete', calculateVariance(parseMoney(get(formData,'totalCost', 'current')), parseMoney(get(formData,'totalCost', 'baseline'))));",
+            "calculation": "get(formData, 'anyChangeToDescription', 'confirmation') === 'Yes' ? set(formData, 'currentFundingStackDescription', get(formData, 'anyChangeToDescription', 'updatedFundingStack')) : set(formData, 'currentFundingStackDescription', get(formData, 'descriptionOfFundingStack')); set(formData['totalCost'], 'percentComplete', calculateVariance(parseMoney(get(formData,'totalCost', 'current')), parseMoney(get(formData,'totalCost', 'baseline'))));",
```
faissalMT: There shouldn't be any need to check for the confirmation to do the calculation, the calculation won't be shown if the relevant object isn't shown.

elcorbs: I pull this funding stack description in from this field in the next return, using the source key, so I want the correct description to be in the field whether it's shown or not.

PR #399, opened 2019-01-25
Corrects migration 14 to cope with incomplete data

faissalMT:


spec/unit/database_administrator/migration_14_spec.rb
```diff
@@ -0,0 +1,112 @@
+# frozen_string_literal: true
+
+fdescribe 'Migration 14' do
```
ccp92:
```suggestion
describe 'Migration 14' do
```

PR #404, opened 2019-01-28
Add edge cases for when statutory consent fields are null

ccp92:


lib/ui/use_case/convert_ui_hif_project.rb
```diff
@@ -100,10 +100,22 @@ def convert_infrastructures
         unless infrastructure[:planningStatus][:statutoryConsents].nil?
           unless infrastructure[:planningStatus][:statutoryConsents][:consents].nil?
             consents = infrastructure[:planningStatus][:statutoryConsents][:consents].map do |consent|
-              {
-                detailsOfConsent: consent[:detailsOfConsent],
-                targetDateToBeMet: consent[:targetDateToBeMet]
-              }
+              if consent[:detailsOfConsent].nil? && consent[:targetDateToBeMet].nil?
```
faissalMT: I would write these by creating a hash and then adding the relevant items conditionally as we've done elsewhere, rather than creating unique hashes for every possibility.

PR #408, opened 2019-01-31
Resolve bug with return submission

ccp92:
Previous changes to error message creation caused issues with array items.
Add edge case where message is an integer (i.e. array level identifier).


lib/common/domain/template.rb
```diff
@@ -50,6 +50,7 @@ def get_dependency_failed_path_names(error, project_data)
       error.each do |message|
         next if message.to_s.include? "is not a member of"
         next if message.data.nil?
+        next if message.data.is_a? Integer
```
faissalMT: How about splitting this into a method called `message_is_array_index`?

PR #410, opened 2019-01-31
PCS

faissalMT:
What - Modify PCS interface to comply with the actual API. I have created the use case `populate_baseline` which can be used anywhere where a `find_project` is currently used. `bid_id` is a column that has been added to the projects domain object and database table. `bid_id`s look like `HIF/MV/5355` and when passed to the PCS API are percent encoded, which may catch you out when stubbing them.


danielburnley:
Is there a reason we decided to reuse the API key the user authenticates with as opposed to using generating a JWT token with a shared key for each request to PCS?

Concerned that we can now no longer revoke all keys by invalidating the secret as it'll break PCS integration. Feels like two separate keys rather than one - so interested to see if there was a specific reason we chose to reuse them

faissalMT:
Now that you mention it, no, there isn't a good reason why that's the case. We could start using the current secret for PCS and create a new secret for future users? That way we don't need to change it on the PCS end (because there are other systems authenticating against PCS).

danielburnley:
Yeah i'd go for that - worth keeping them separate :ok_hand:

spec/acceptance/local_authority/authorise_user_with_project_spec.rb
```diff
@@ -28,6 +29,21 @@
       api_key = get_use_case(:create_api_key).execute(projects: [1], email: 'cat@cathouse.com', role: 'HomesEngland')[:api_key]
       expect(get_use_case(:check_api_key).execute(api_key: api_key, project_id: 1)).to eq(valid: true, email: 'cat@cathouse.com', role: 'HomesEngland')
     end
+
+    it 'should create a valid pcs key for project 1' do
+      api_key = api_key = get_use_case(:create_api_key).execute(projects: [1], email: 'cat@cathouse.com', role: 'Homes England')[:api_key]
```
elcorbs:
```suggestion
      api_key = get_use_case(:create_api_key).execute(projects: [1], email: 'cat@cathouse.com', role: 'Homes England')[:api_key]
```

PR #430, opened 2019-02-15
Adapt to allow project creation and adding users to projects

elcorbs:
I've had to add different webroutes for the admin access so we can still add projects and people like we did before, but with a different url. But if any ideas how to do it with the same webroute, that would be good to know.

### DIFF
--- a/lib/ui/gateway/in_memory_new_project.rb
+++ b/lib/ui/gateway/in_memory_new_project.rb
@@ -1,0 +1,20 @@
+class UI::Gateway::InMemoryNewProject
+  def find_by(type:)
+    if type == 'hif'
+      create_project('mvf')
+    elsif type == 'ac'
+      create_project('ac')      
+    else
+      return nil
+    end
+  end
+
+  private
+
+  def create_project(type)
+    JSON.parse(
+      File.open("#{__dir__}/blank_projects/#{type}.json", 'r').read,
+      symbolize_names: true
+    )
+  end
+end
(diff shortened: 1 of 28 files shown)


lib/ui/gateway/in_memory_new_project.rb
```diff
@@ -0,0 +1,20 @@
...
+      return nil
+    end
+  end
+
+  private
+
+  def create_project(type)
+    JSON.parse(
+      File.open("#{__dir__}/blank_projects/#{type}.json", 'r').read,
+      symbolize_names: true
```
ccp92: Need to call File.close() after calling File.open() in a function.

elcorbs: I micked the same way that we have done this in in_memory_returns_schema and in_memory_project_schema

elcorbs: was it causing as error? should we be changing all of them?

elcorbs: @ccp92

ccp92: It's not causing an error. Was speaking to @danielburnley as I wasn't sure but basically unless the file open is called in a block (in which case it's implicitly closed at the end of the block) we should be calling the close method.
It can lead to memory errors down the line where too many files are open.

But if it's occurring in multiple places, maybe just merge this PR for now and then we can add the task to the backlog and get all of them in one go at a later date?

PR #432, opened 2019-02-18
Future funding infrastructure list

faissalMT:


lib/delivery_mechanism/controllers/get_project_infrastructures.rb
```diff
@@ -0,0 +1,10 @@
+require_relative '../web_routes.rb'
+
+DeliveryMechanism::WebRoutes.get '/project/:id/infrastructures' do
```
danielburnley: Is infrastructures generic across all projects - or is it specific to a certain subset? :eyes: Might be worth discussing between yourselves how to approach this going forward

lib/local_authority/use_case/get_infrastructures.rb
```diff
@@ -0,0 +1,17 @@
...
+  def initialize(find_project:)
+    @find_project = find_project
+  end
+
+  def execute(project_id:)
+    project = @find_project.execute(id: project_id)
+
+    infrastructures = project.dig(:data, :infrastructures).map do |infrastructure|
+      {
+        information: infrastructure[:information]
```
elcorbs: Why doesn't this return everything in infrastructures.

spec/unit/ui/use_case/convert_core_ff_project_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    JSON.parse(
+      File.open("#{__dir__}/../../../fixtures/ff_baseline_core.json").read,
+      symbolize_names: true
+    )
+  end
+
+  let(:ui_data_project) do
+    JSON.parse(
+      File.open("#{__dir__}/../../../fixtures/ff_baseline_ui.json").read,
+      symbolize_names: true
```
elcorbs: Chris pointed out in my last PR that we should be closing these files, so I've moved the rest of the File.open bits into do blocks

lib/delivery_mechanism/controllers/get_project_infrastructures.rb
```diff
@@ -0,0 +1,10 @@
...
+  guard_access env, params, request do |_|
+    infrastructures = @dependency_factory.get_use_case(
+      :get_infrastructures
+    ).execute(project_id: params['id'].to_i)
+    infrastructures.to_json
```
elcorbs: I think this should be sending the schema for infrastructure as well. or were you thinking we would still use find by to get the schema?

faissalMT: Yes I was under the impression we were only using get_infrastructures to create the dropdown items

elcorbs: Yeah ok, I was thinking we would use it to display them as well, but it probably makes more sense to do that using the normal find_by route.

PR #434, opened 2019-02-19
Resolve error when saving or submitting projects and returns

ccp92:


lib/ui/use_case/get_schema_for_return.rb
```diff
@@ -4,11 +4,11 @@ def initialize(get_return:, return_template:)
     @return_template = return_template
   end
 
-  def execute(return_id:)
-    type = @get_return.execute(id: return_id)[:type]
+  def execute(return_id:, api_key:)
+    type = @get_return.execute(id: return_id, api_key: api_key)[:type]
```
danielburnley: `api_key` seems like a delivery mechanism concern - This will cause your acceptance tests for the core logic to now be tied to being delivered through an authenticated API. 

I'd expect `get_return` to take the data you pull from the API Key rather than the raw API key itself, as it makes it much harder to reuse the `get_return` use case now it's tied to the API key

danielburnley: I realise solving this might be outside the scope of this PR - but it's worth highlighting that passing the API key through the system couples the business logic to the delivery mechanism and makes it much less portable as you'll now always need an API Key with the correct data stored within it, which makes the tests much harder to write also

ccp92: As mentioned in slack, I'm now realising that the api_key the get_project and get_return usecases was expecting were probably the PCS api information. Will check with Faissal what needs to be done.

lib/ui/use_case/get_schema_for_return.rb
```diff
@@ -4,11 +4,11 @@ def initialize(get_return:, return_template:)
     @return_template = return_template
   end
 
-  def execute(return_id:)
-    type = @get_return.execute(id: return_id)[:type]
+  def execute(return_id:, api_key:)
```
elcorbs: We could just pass the type through here, as the only reason this calls get return is to get the type. Also in the webroute where this is used it gets the return inc. the type just beforehand.

spec/web_routes/update_project_spec.rb
```diff
@@ -45,6 +47,16 @@
       double(new: double(execute: {valid: true}))
     )
 
+    stub_const(
```
faissalMT: You can use `stub_instances(LocalAuthority::UseCase::ApiToPcsKey, api_to_pcs_key_spy)` instead. There also doesn't seem to be a test making sure that this is being called correctly.

spec/web_routes/update_project_spec.rb
```diff
@@ -107,5 +107,22 @@
         )
       )
     end
+
+    it 'should check the validity of the api key' do
```
faissalMT: These could easily be slimed and so need triangulating.

PR #442, opened 2019-02-21
Store claims

faissalMT:


lib/local_authority/gateway/sequel_claim.rb
```diff
@@ -0,0 +1,35 @@
+class LocalAuthority::Gateway::SequelClaim
+  def initialize(database:)
+    @database = database
+  end
+
+  def create(claim)
+    @database[:claims].insert(
+      project_id: claim.project_id,
+      type: claim.type,
+      status: claim.status,
```
elcorbs: Should bid id not be here aswell?

faissalMT: Bid id comes from the projects table, we get it when we join with projects https://github.com/homes-england/monitor-api/blob/c3c3700d1359eb051c8d64389699bee31f101ad6/lib/local_authority/gateway/sequel_claim.rb#L20

spec/unit/ui/use_case/convert_ui_ac_return_spec.rb
```diff
@@ -1,5 +1,5 @@
 describe UI::UseCase::ConvertUIACReturn do
-  let(:return_to_convert) do
+  let(:claim_to_convert) do
```
elcorbs: This should still be return - or the file name/ class it's testing is wrong

lib/local_authority/gateway/sequel_claim.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    LocalAuthority::Domain::Claim.new.tap do |claim|
+      claim.project_id = row[:project_id]
+      claim.type = row[:type]
+      claim.bid_id = row[:bid_id]
+      claim.status = row[:claims_status]
+      claim.data = Common::DeepSymbolizeKeys.to_symbolized_hash(row[:claims_data].to_h)
+    end
+  end
+
```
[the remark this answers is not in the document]
faissalMT: Yes.

PR #449, opened 2019-03-01
Forward Funding Procurement Tab

ccp92:
Add Procurement tab to Forward Funding

This PR also splits the tabs for FF into seperate files to make it easier to manage them going forwards.


danielburnley:
(summarising face to face chat) 

It's probably worth looking at splitting these tabs into separate files for now to save us pain down the line

lib/ui/gateway/schemas/ff_project.json
```diff
@@ -1395,6 +1395,110 @@
...
+            }
+          }
+        },
+        "summaryOfProcurement": {
+          "type": "object",
+          "title": "Summary of key procurements",
+          "properties": {
+            "procurements": {
+              "type": "array",
+              "title": "",
```
elcorbs: Might be good to put a title here to break it up, otherwise it looks weird if there are two after to each other. 
screen shot 2019-03-04 at 09 10 36

PR #451, opened 2019-03-04
Add Demolition and Remediation Tab

ccp92:
WHAT:
New tab (screenshot below)
Removed ruby conversion to JSON in ff_template

screenshot-localhost-3000-2019 03 04-12-41-45


lib/homes_england/builder/template/templates/ff_template.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class HomesEngland::Builder::Template::Templates::FFTemplate
```
danielburnley: Is there a reason we're defining this json structure in ruby :eyes: 

Feels like we should do better this time and also look at storing this data in json because it is at the end of the day json, and editing it in ruby is a nightmare

danielburnley: I realise the old structures are in Ruby - but looking to change that going forward

PR #452, opened 2019-03-04
Add milestones tab

ccp92:
screenshot-localhost-3000-2019 03 04-14-28-08


lib/ui/gateway/schemas/ff/project/milestones.json
```diff
@@ -0,0 +1,30 @@
+{
+  "type": "object",
+  "title": "Key Project Milestones",
+  "properties": {
+    "milestones": {
+      "type": "array",
+      "title": "Milestones",
+      "addable": true,
+      "items": {
+        "type": "object",
```
elcorbs: Same as before, I think it needs a title to break up the milestones. Maybe remove the title on the array and put one here instead ? 
screen shot 2019-03-05 at 10 04 16

PR #453, opened 2019-03-04
Add template builder for split out schemas

danielburnley:
What:
- Adds a template builder to allow for building up sections of a larger more complex schema

Currently only supports adding sections in the properties -will need to be extended as we go to split up the current schemas to add additional top level items

### DIFF
--- a/lib/ui/builder/template.rb
+++ b/lib/ui/builder/template.rb
@@ -11,7 +11,7 @@ class UI::Builder::Template
 
   def add_section(name:, file_name:)
     File.open("#{@path}/#{file_name}") do |f|
-      @schema[:properties][name.to_sym] = JSON.parse(f.read, symbolize_names: true)
+      @schema[:properties][name] = JSON.parse(f.read, symbolize_names: true)
     end
   end
 
--- a/lib/ui/builder/template.rb
+++ b/lib/ui/builder/template.rb
@@ -1,0 +1,23 @@
+class UI::Builder::Template
+  def initialize(path:, title:)
+    @path = path
+    @schema = {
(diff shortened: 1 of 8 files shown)


lib/ui/builder/template.rb
```diff
@@ -0,0 +1,23 @@
...
+      '$schema' => 'http://json-schema.org/draft-07/schema',
+      title: title,
+      type: 'object',
+      properties: {}
+    }
+  end
+
+  def add_section(name:, file_name:)
+    File.open("#{@path}/#{file_name}") do |f|
+      @schema[:properties][name.to_sym] = JSON.parse(f.read, symbolize_names: true)
```
ccp92: Could we not just pass the symbol in as an argument directly rather than converting a string into a symbol here?

danielburnley: Nope - no reason. Have changed to avoid meaningless conversion

lib/ui/builder/template.rb
```diff
@@ -0,0 +1,23 @@
...
+    @path = path
+    @schema = {
+      '$schema' => 'http://json-schema.org/draft-07/schema',
+      title: title,
+      type: 'object',
+      properties: {}
+    }
+  end
+
+  def add_section(name:, file_name:)
```
ccp92: Sorry, this is probably super nit-picky, but rather than `name`:, maybe `section:` would be a better variable name?
That way it's explicit about what the name is referring to as name is a bit washy?

danielburnley: :+1: sounds good to me

PR #454, opened 2019-03-04
Add risks tab to forward funding baseline

ccp92:
screenshot-localhost-3000-2019 03 04-16-05-59

### DIFF
--- a/lib/homes_england/builder/template/templates/ff_template.json
+++ b/lib/homes_england/builder/template/templates/ff_template.json
@@ -1467,7 +1467,12 @@
                   "status": {
                     "type": "string",
                     "title": "Status at Baseline",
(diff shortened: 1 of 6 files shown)


lib/homes_england/builder/template/templates/ff_template.json
```diff
@@ -1631,6 +1636,41 @@
...
+      "type": "object",
+      "title": "Risks",
+      "properties": {
+        "risks": {
+          "type": "array",
+          "addable": true,
+          "title": "",
+          "items": {
+            "type": "object",
+            "properties": {
```
elcorbs: Same thing about a title here? 
screen shot 2019-03-05 at 10 07 05

PR #458, opened 2019-03-05
Add hif recovery tab to forward funding

ccp92:
screenshot-localhost-3000-2019 03 05-12-03-45

### DIFF
--- a/lib/homes_england/builder/template/templates/ff_template.json
+++ b/lib/homes_england/builder/template/templates/ff_template.json
@@ -1904,8 +1904,7 @@
         }
       }
     },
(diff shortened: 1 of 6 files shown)


lib/homes_england/builder/template/templates/ff_template.json
```diff
@@ -1903,6 +1903,55 @@
           }
         }
       }
+    },
+    "hifRecovery": {
+      "$schema": "http://json-schema.org/draft-07/schema",
```
elcorbs: Does this have to be on every tab?

PR #460, opened 2019-03-05
Begin adding validation for claims

faissalMT:


elcorbs:
Do you reckon it's worth using the same validate use case for projects/ returns/ claims? They are all the same, it seems weird to repeat it.

ccp92:
@elcorbs Spoken to Dan regarding this, and although there is some duplication of code between the three validation use cases, there could be deviation between the intent of the three going forwards.
(i.e. we may need to validate different things in different ways).

PR #462, opened 2019-03-05
Add outputs tab along with ability to add shared data

ccp92:
screenshot-localhost-3000-2019 03 05-14-48-40


lib/ui/gateway/schemas/ff/project/outputs.json
```diff
@@ -0,0 +1,44 @@
...
+          "title": "Forecast",
+          "type": "array",
+          "periods": true,
+          "addable": true,
+          "items": {
+            "title": "Forecast",
+            "type": "object",
+            "properties": {
+              "period": { "type": "string", "title": "Period" },
+              "target": {
```
elcorbs: This should probably be called housingStarts to avoid confusion

PR #465, opened 2019-03-05
Add version numbers to baselines

elcorbs:
Also, adds a baseline version number to each return.


spec/acceptance/homes_england/amend_project_spec.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+require 'rspec'
+require_relative '../shared_context/dependency_factory'
+
+describe 'Amending a project' do
```
danielburnley: The use case is called "Amend baseline" - feels like we're being inconsistent with our naming. Might be worth thinking about renaming projects to baselines, but, that feels like a lot of change. 

For now i'd consider calling the use case `Amend_project` so it's consistent, and see about renaming things after this PR

db/migrations/18_add_version.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:projects) do
```
danielburnley: How do i get the data for the previous baseline? If I want to view the baseline a project was created against, can I see that? Or do we overwrite it

spec/acceptance/homes_england/amend_project_spec.rb
elcorbs: We went with amend baseline ad it seems more in line with what HE call it. Also if we are detaching baseline data from a project then does it make sense to call it amend project?

db/migrations/18_add_version.rb
elcorbs: We are holding off making this decision until we know more about how the user plans to use this functionality.

spec/acceptance/homes_england/amend_project_spec.rb
danielburnley: True, true - we do now have the chance to separate the term "project" and "baseline"

db/migrations/18_add_version.rb
danielburnley: :+1:

db/migrations/18_add_version.rb
```diff
@@ -0,0 +1,9 @@
...
+      add_column :version, Integer, default: 1
```
danielburnley: Still wonder if version is needed here - Given the version will also have to be stored against the baseline revisions. Maybe `latest_version` could be more accurate, if you wanted to be able to use it to speed up searching for baseline revisions but otherwise it's not really required here IMO

danielburnley: This works for now at least but feels like this is gonna change very swiftly as a result of what we discussed yesterday

elcorbs: I have already changed this - but i've branch off this branch. I can just do a PR at the end but was worried it would be too big

PR #468, opened 2019-03-06
Get base claims 

elcorbs:


docker-compose.yml
```diff
@@ -26,7 +26,6 @@ services:
       EMAIL_WHITELIST: 'test@test.test'
       GOV_NOTIFY_API_KEY: 'cafe-cafecafe-cafe-cafe-cafe-cafecafecafe-cafecafe-cafe-cafe-cafe-cafecafecafe'
       HMAC_SECRET: 'E'
-      PCS: 'yes'
```
ccp92: This needs to be re-added

ccp92:
```suggestion
      PCS: 'yes'
```

elcorbs: but it doesn't work?

PR #473, opened 2019-03-11
Add ff funding to project create

ccp92:
Add blank ff project template to create from front-end.
Requires https://github.com/homes-england/monitor-frontend/pull/308

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -53,6 +53,7 @@ services:
       FF_FUNDING_PACKAGE_TAB: 'yes'
       FF_WIDER_SCHEME_TAB: 'yes'
       NO_LOGS: 'yes'
+      FF_CREATION: 'yes'
 
   notify_simulator:
     build: .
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -52,6 +52,8 @@ services:
       FF_RECOVERY_TAB: 'yes'
       FF_FUNDING_PACKAGE_TAB: 'yes'
       FF_WIDER_SCHEME_TAB: 'yes'
+      NO_LOGS: 'yes'
```
```suggestion
```

ccp92: Dan identified that the logging is the cause of slowdown when trying to create a return.
As such, should disable it by default and turn it back on locally as and when it's required.
(The flag has been turned on in staging also).

ccp92:
> The Infrastructure funding package tab crashes

This was due to the blank data being used to create the project not containing and array for that tab. That's now been resolved so creating a project through the portal will no longer crash that tab.

PR #491, opened 2019-03-26
Add endpoint to get all claims

elcorbs:


spec/web_routes/get_claims_spec.rb
```diff
@@ -0,0 +1,106 @@
+# frozen_string_literal: true
+
+require 'rspec'
+require_relative 'delivery_mechanism_spec_helper'
+
+describe 'Getting return history' do
+  let(:get_claims_spy) { spy(execute: returned_hashes) }
+
+  before do
+    stub_const(
```
faissalMT: We can use `stub_instances()` now, i.e: `stub_instances(UI::UseCase::GetClaims, get_claims_spy)`

PR #493, opened 2019-03-26
Remove unnecessary questions when N/A is selected

elcorbs:


lib/ui/gateway/schemas/hif_project.json
```diff
@@ -2060,14 +2060,225 @@
...
+                      }
+                    },
+                    {
+                      "properties": {
+                        "programmes": { "enum": ["Accelerated Construction"] },
+                        "description": {
+                          "type": "string",
+                          "title": "Description"
+                        },
+                        "pcsRef": {
```
faissalMT: Since PCS is being input on project creation, it's best we just remove this field for now.

PR #523, opened 2019-04-04
Add ability to change user role

faissalMT:


spec/acceptance/local_authority/update_user_role_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    get_use_case(:ui_create_project).execute(
+      type: 'hif',
+      name: 'cat sanctuary',
+      baseline: nil,
+      bid_id: '12345'
+    )[:id]
+  end
+  include_context 'dependency factory'
+
+  before do
```
danielburnley: To avoid this causing issues with other tests (not likely with this env vars but still good practice) it's worth doing something like: 

```ruby
let(:env_before) { ENV['ENV_NAME'] }

before do
    env_before
    ENV['ENV_NAME'] = 'Some animal noise'
end

after do
     ENV['ENV_NAME'] = env_before
end
```

This prevents overwriting things for further tests with environment variables, especially useful when dealing with toggle features

spec/web_routes/update_user_spec.rb
```diff
@@ -0,0 +1,93 @@
+# frozen_string_literal: true
+
+require 'rspec'
+require_relative 'delivery_mechanism_spec_helper'
+
+describe 'Admin changing user role' do
+  ENV['ADMIN_HTTP_API_KEY'] = 'supersecret'
```
danielburnley: This should probably be in some kind of before/after thing also

PR #525, opened 2019-04-04
Validation errors

elcorbs:
- Remove empty string before validating
- Discard irrelevant error messages within dependencies


lib/common/domain/template.rb
```diff
@@ -5,12 +5,28 @@ class Common::Domain::Template
...
 
   private
 
+  def data_present?(value)
+    return false if value.nil?
+    return true unless value.is_a?(String)
+    return !value.strip.empty?
+  end
+  
+  def compact_data(data)
```
danielburnley: Part of me wonders if this should be its own use case - it feels like it's sanitising data from the frontend :thinking: If by the time it was being validated the data was sanitised, it would be pretty good

We could then use it in any place we take in data from the frontend (claims, returns, baselines, etc)

danielburnley: Or, alternatively, we could make it a refinement on the Hash class so it could be used in multiple places (i think I prefer this one). 

```ruby
module Common
  module Refinements
    module HashCompacting
      refine ::Hash do
        def deep_compact
           ### Method
        end
      end
    end
  end
end
```

elcorbs: it needs to be behind the convertors because they add nils to things

spec/unit/ui/use_case/convert_core_hif_project_spec.rb
```diff
@@ -45,4 +45,18 @@
     converted_project = described_class.new.execute(project_data: nil_data_to_convert)
     expect(converted_project).to eq(returned_empty_project)
   end
+
+  # let(:invalid_project) do
```
danielburnley: :eyes: can we delete this if it's not used

PR #530, opened 2019-04-08
Add support for minimum array item validation

danielburnley:


lib/common/domain/template.rb
```diff
@@ -80,7 +82,7 @@ def get_required_failed_node_names(error)
 
   def get_supported_errors(aggregated_errors)
     aggregated_errors.errors.select do |error|
-      %i[required_failed one_of_failed].include? error.type
+      %i[required_failed one_of_failed min_items_failed].include? error.type
```
faissalMT: I feel like this reads really badly to the point where I'd be inclined to monkeypatch  
`error.type.in %i[required_failed one_of_failed min_items_failed]`  
I don't know how we feel about monkeypatching though, and if this is a rare occurrence it might not be worth it.

danielburnley: Yah lets avoid monkey patching, if we have to add another though it's worth looking at cleaning it up. 

Your suggestion does look much nicer though

PR #540, opened 2019-04-15
Create project overview use case

danielburnley:
To be wired up to a controller in a subsequent PR

### DIFF
--- a/lib/delivery_mechanism/controllers/project_overview.rb
+++ b/lib/delivery_mechanism/controllers/project_overview.rb
@@ -1,0 +1,22 @@
+require_relative '../web_routes.rb'
+
+DeliveryMechanism::WebRoutes.get '/project/:id/overview' do
+  guard_access env, params, request do |_|
+    project_overview = @dependency_factory.get_use_case(:get_project_overview).execute(
+      id: params[:id].to_i
+    )
+    return 404 unless project_overview
+
+    response.body = {
+      name: project_overview[:name],
+      status: project_overview[:status],
+      type: project_overview[:type],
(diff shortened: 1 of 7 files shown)


spec/unit/homes_england/use_case/get_project_overview_spec.rb
```diff
@@ -0,0 +1,239 @@
+describe HomesEngland::UseCase::GetProjectOverview, focus: true do
```
faissalMT:
```suggestion
describe HomesEngland::UseCase::GetProjectOverview do
```
Also sidenote: you can use fdescribe instead or just :focus

lib/delivery_mechanism/controllers/project_overview.rb
```diff
@@ -0,0 +1,22 @@
+require_relative '../web_routes.rb'
+
+DeliveryMechanism::WebRoutes.get '/project/:id/overview' do
+  guard_access env, params, request do |_|
```
faissalMT:
```suggestion
  guard_access env, params, request do
```

spec/web_routes/project_overview_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+
+require 'rspec'
+require_relative 'delivery_mechanism_spec_helper'
+
+describe 'Getting the project overview' do
+  let(:project_overview_spy) { spy(execute: found_overview) }
+  let(:found_overview) { nil }
+
+  before do
+    stub_const(
```
faissalMT: We're using a wrapper function called `stub_instances()` for most of these now

PR #552, opened 2019-04-25
Add support for pattern in one of

danielburnley:


spec/unit/ui/use_case/validate_project_spec.rb
```diff
@@ -1278,6 +1278,65 @@ def execute(argument)
   end
 
   context 'shows pattern validation' do
+    context 'Outside of a one of' do
+      it_should_behave_like 'required field validation'
+      let(:template) do
+        Common::Domain::Template.new.tap do |p|
```
faissalMT: We should probably start avoiding using `p` as a variable since it shadows the `p` function

PR #560, opened 2019-05-03
Move web routes to controllers

RobertM905:
Moved some web routes into individual controllers to make future changes easier.


lib/delivery_mechanism/controllers/get_base_return.rb
```diff
@@ -0,0 +1,19 @@
+require_relative '../web_routes.rb'
+
+DeliveryMechanism::WebRoutes.get '/project/:id/return' do
+  guard_access env, params, request do |_request_hash|
```
faissalMT: Since `|_request_hash|` isn't being used you can just remove it, there's a few other places where this is still there too.

PR #581, opened 2019-05-15
Refactor validation acceptance specs

faissalMT:


spec/acceptance/ui/hif_validate_baseline_spec.rb
```diff
@@ -36,9 +47,15 @@
...
-      expect(valid_project[:pretty_invalid_paths]).to eq(PRETTY_INVALID_PATH)
+      expect(validated_project[:valid]).to eq(false)
+      expect(validated_project[:invalid_paths]).to eq(invalid_path)
+      expect(validated_project[:pretty_invalid_paths]).to eq(pretty_invalid_path)
+    end
+
+    it 'should return invalid if fails validation' do
+      given_an_invalid_project
+      when_validated
+      then_the_correct_fields_are_marked_invalid
```
danielburnley: I think we should look at checking every invalid path (even if that is a _massive_ list) if we're looking at these acceptance specs. 

We can at least define them in a fixture file - but it would be good now that they're relatively settled that we can then start being sure that no change to the validation will change what we know to be a valid path (other than actually explicitly changing the schema)

danielburnley: This feedback applies to the other validation specs FYI

They could maybe be merged into one which is `hif_validation_spec` or something which validates all the things with their base equivalents (e.g. base claim, base return, initial baseline)

